PR #2, opened 2021-08-06
Slightly adjust map icons and merchant item descriptions in ultrawide

CodingSquirrel:
Map in 21:9 should exactly match the 16:9 positioning now. Also more accurately match the positioning for merchant item description text.

It bugged me slightly, so I figured I'd fix it myself and save you the trouble :)


nipkownix:
Nice! Thank you a lot. It would be quite hard for me to perfectly adjust this without having an ultrawide monitor.

PR #7, opened 2021-11-03
Improved `RestorePickupTransparency` patch

emoose:
Improved the `RestorePickupTransparency` patch as described at https://github.com/nipkownix/re4_tweaks/issues/5#issuecomment-957993252

I'm not really sure what to look for to know if lighting is fixed or not though, do you have any way to check that?

Only tested this with 1.1.0, but pretty sure the sig should catch 1.0.6 too.


nipkownix:
Looks like it is working flawlessly! Thanks!

PR #14, opened 2021-11-08
Add `WindowBorderless`, `WindowPositionX` & `WindowPositionY` settings, fixes #13

### LINKED ISSUE #13: Feature request: allow setting window position in windowed mode
Would be nice if window position could be set in winmm.ini or something, as I have my taskbar-on-top the game seems to have a bad habit of drawing the screen behind it, so the first 50 pixels or so have my taskbar icons on top of it :/

I think window position can be set when CreateWindow is used, might be able to specify borderless-window by changing CreateWindow params too, remember doing something like that for sh2proxy anyway.

(or maybe there's a way to make the game move itself after the taskbar automatically - in other apps that render behind taskbar, usually I can get them to move by just dragging bottom-right & resizing the window, doing that usually gets Windows to automatically move it, wonder if that works programmatically too)

Even better would be some way of saving where the user dragged the window to, Windows might be able to handle that itself, but I think game does something at startup that moves to 0,0 :/

E: wow, it took 4 different patches to completely stop game from moving the window:
```C
  SafeWrite(GameAddress + 0x54B7D5, uint8_t(0x90), 6);
  SafeWrite(GameAddress + 0x53A219, uint8_t(0x90), 6);
  SafeWrite(GameAddress + 0x53A226, uint8_t(0x90), 6);
  SafeWrite(GameAddress + 0x53A243, uint8_t(0x90), 6);
```
I think only the last one would need to be hooked there, but maybe nopping all those + hooking CreateWindowExA could be a nice idea, could set the location once so the window won't have to move/snap itself to a position after initing like it currently does.

E: got something working, with borderless windows too :) gonna try making some sigs for it now.

emoose:
Tested with 1.1.0 & 1.0.6, hopefully sig should match 1.0.6 JP too.

Might be nice if the INI could also be updated when user moves the window around, via the WndProc hook maybe, though it seems RE4 doesn't always let you move the window for some reason...

(also I still wonder if there's any way to make Windows automatically move it out of the way of taskbar, didn't have any luck with that yet though)

### DIFF
--- a/dllmain/dllmain.cpp
+++ b/dllmain/dllmain.cpp
@@ -51,6 +51,10 @@ bool bFixBlurryImage;
 bool bDisableFilmGrain;
 bool bIgnoreFPSWarning;
 bool bMemOptimi;
+bool bWindowBorderless;
+
+int iWindowPositionX;
+int iWindowPositionY;
 
 uintptr_t jmpAddrChangedRes;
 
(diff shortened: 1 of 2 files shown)


emoose:
Oh I forgot to add updated INI, 1sec

emoose:
Oh hmm, seems this might stop resolution changes when in windowed mode though... I'll try fixing it soon.

emoose:
That should fix it hopefully, now also stops game from reapplying window borders when resolution changes (if bWindowBorderless is set)

nipkownix:
The position is now automatically saved as of d88f18500.

PR #16, opened 2021-11-09
Add hooks to make DoF blur closer to GC, add `EnableGCBlur` setting

emoose:
Added blur fixes from https://github.com/nipkownix/re4_tweaks/issues/5#issuecomment-963766989

Seems to be working across both 1.1.0 & 1.0.6, haven't tried 1.0.6 JP though. Should hopefully match up with the DLL that I posted in that thread, but haven't tried with that many cutscenes yet.

Not sure if there might be a better way to handle the sigs - couldn't sig-scan for the GX funcs themselves since they're all so similar to each other, so had to scan for code that called them instead & find the offset based on the `call` insn, luckily didn't need too much code to handle that though.

I figured `GCBlur` would probably sound better than `DoFBlur` here, but feel free to rename the setting/vars/etc as you like of course!

E: might need to look into increasing the focus-time for binoculars/scope, seems much faster than GC when at 60FPS at least...

### DIFF
--- a/dllmain/dllmain.cpp
+++ b/dllmain/dllmain.cpp
@@ -59,6 +59,7 @@ bool bUseMemcpy;
 bool bIgnoreFPSWarning;
 bool bMemOptimi;
 bool bWindowBorderless;
+bool bEnableGCBlur;
 
 uintptr_t jmpAddrChangedRes;
 uintptr_t* ptrYMovieResFromFile;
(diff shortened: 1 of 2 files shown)


nipkownix:
Looks pretty good! Merged.
Congrats for making this happen!

Keith94:
Can you disable blur completely with this patch?

emoose:
The vanilla game mostly has all the blur disabled already really, if you notice blur in the village section that might be caused by the early HD project release, other areas should be blur free though IIRC.

PR #17, opened 2021-11-11
Add `SkipIntroLogos` option to skip the Capcom etc logos at startup

emoose:
This allows skipping most of the startup logo sequence, it's not entirely skipped as it seems that sequence is needed to load certain things in for the main menu etc, but this method is the best way I've found so far to skip the longer logo sections of it.

Tested with 1.1.0 & 1.0.6, seems to work fine - comes in very useful when you're constantly restarting the game trying to track down an annoying crash :P


nipkownix:
Looks pretty good to me! Nice job!

PR #18, opened 2021-11-12
Add support for 1.0.6 debug build

emoose:
Looks like 1.0.6 debug build works fine with most of the sigs here, only ones that it had trouble with were

```C
	// HealthBase pointer
	auto pattern = hook::pattern("A1 ? ? ? ?  83 C0 60  6A 10  50  E8");
	HealthBase = *pattern.count(1).get(0).get<uint32_t*>(1);
```
and
```C
	// Inventory screen mem
	pattern = hook::pattern("A1 ? ? ? ? 56 D9 58 ? FF 15 ? ? ? ? 50 FF D7 8B 8D ? ? ? ? 8B 95 ? ? ? ? 89 0D");
	ptrpG = injector::ReadMemory<uint32_t*>(pattern.count(1).get(0).get<uint32_t*>(1), true);
```

It turns out both of these actually retrieve the same value too, the pG pointer (0xC06F3C in 1.1.0)
(also my debug menu stuff actually searches for pG yet again as well, so that'd make 3 patterns searching for the same thing 😄)

The pattern I used for debug menu seemed to work across both 1.0.6/1.1.0 and the debug build though, so I merged those two and updated it with that, after that it seemed the rest can be found fine (I still need to check that everything still works though - will update this PR later on)

Haven't added anything to differentiate 1.0.6 non-debug & debug yet though, so `game_version` will still stay `1.0.6` for both, will probably add something for this in the debug-menu stuff so it only patches what's needed (afaik debug build just needs the `roomInfo.dat` stuff to fix `AREA JUMP`, and maybe some code to slow the menus down so they're actually usable)

### DIFF
--- a/dllmain/dllmain.cpp
+++ b/dllmain/dllmain.cpp
@@ -767,9 +767,8 @@ void HandleLimits()
 	injector::WriteMemory<int>(ptrp_MemPool_SubScreen16, 0x000000, true);
 	injector::WriteMemory<int>(ptrp_MemPool_SubScreen17, 0x000000, true);
 
-	#define SS_MEM_OFFSET 0x1CB5400 // SubScreenAramRead adds this offset to the memory addr, could probably be patched out, but meh
-	injector::WriteMemory<int>(ptrg_MemPool_SubScreen1, (uint32_t)(sizeof(g_MemPool_SubScreen) - SS_MEM_OFFSET), true); // SubScreenExec MemorySwap size
-	injector::WriteMemory<int>(ptrg_MemPool_SubScreen2, (uint32_t)(sizeof(g_MemPool_SubScreen) - SS_MEM_OFFSET), true); // SubScreenExitCore MemorySwap size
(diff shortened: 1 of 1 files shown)


emoose:
Hmm weird, had a crash after trying to open inventory (with original ss_pzzl), and now every time I try loading a save it crashes inside some kind of memcpy func.

Looks like 0x5AA9D0 (1.0.6 debug) is the caller, not sure what that func is doing though.

emoose:
Ah think I found the issue, the `game_version == "1.0.6"` checks in the subscreen-pattern code changed the offset to patch, but seems the 1.0.6 debug doesn't need the offset changed, just added some code to add `d` to game_version if it's a debug build, so those changes won't happen.

nipkownix:
Oh, I thought there'd be more problems with sigs, to be honest. Thanks for looking into this!

emoose:
Seems like it should be good to go, I haven't been able to test `FixUltraWideAspectRatio` or most of the `FixQTE` patches though, but I'd assume they should work, looks like 1.0.6 debug is pretty close to 1.1.0, so hopefully everything that works on 1.1.0 should work here too.

nipkownix:
Nice! Thank you.

PR #21, opened 2021-11-24
Add `EnableGCScopeBlur` option with Filter0a fixes

emoose:
Adds the Filter0a scope-blur fixes from #5 as part of the `EnableGCBlur` option (not sure if this should maybe be split off from that though, I can imagine some people will see the word DoF in the description and disable it straight away, and miss out on the non-DoF scope blur fix :/)
(E: updated this to split it into a `EnableGCScopeBlur` setting instead)
(E2: also this doesn't affect the infrared scope, IIRC that uses a different Filter to work, will look into that another time)

The Filter0aGXDraw hook is done in a special way, instead of needing to sig-scan & reimplement all the GXInitTexObj etc calls that it makes, we can instead just disable the part of GXDraw that we reimplement, then from our hook call into that neutered GXDraw to handle all that init crap for us - but unfortunately that GXDraw code makes use of ECX for a parameter, which is what made non-debug-builds crash too AFAIK (couldn't find any calling-conventions that match up with it, __fastcall is close but GXDraw cleans stack like __cdecl... guess it's probably a custom convention made by LTO or something)

Luckily wasn't too hard to get around the ECX-parameter by just making a small trampoline function to set ECX for us instead - since our hook doesn't take any ECX parameter we lose what the actual value of it is though, but fortunately it can usually be worked out from `ScreenSizeDivisor` param.

(A better way might be to patch the GXDraw func with a call to our hook code instead, not really sure how to handle passing all the args over to it etc though, but that way would solve the ECX issue...)

Also added some tweaks to the mask-position to try centering it a bit, seems a little better now, but still different from GC - can't really tweak the position any further without also shifting the blurred image too though, but maybe replacing the scope/mask textures could improve it more.

Tested with 1.1.0/1.0.6/1.0.6-debug, seems to work fine 😄


emoose:
Also this made me notice that the crosshairs in the scopes above aren't actually centered properly, the right-side one is closer to middle than left-side is, need to check how those are on GC.

nipkownix:
> (E: updated this to split it into a EnableGCScopeBlur setting instead)

Good call. 

> E: ah GC also has it slightly off center too:

Oh great, I'll never be able to unsee it lol
Nice catch.

> E: tried some things with PC ver but couldn't get it to match with GC,

I'd say a 100% match with GC isn't really necessary, tbh. Its not like the GC version has no flaws :p
Your version looks really good though! Can't say I ever expected to see this effect back on PC. (same with the DoF).
Amazing job! Congrats!
Let me know once you want this merged.

Issue #23, opened 2021-11-27
Rifle reload animation


emoose:
So this is a pretty popular request, I've tried naming & looking into all the sniper/Wep09 related funcs in my IDB, but couldn't really find much to help with it so far.

Looks like animations are stored as FCV format files, usually inside the weapons BIO4/Em/wepXX.udas.lfs file.
Apparently the FCV stores things like frame numbers & counts inside it, but does the EXE have anything to interpolate that to work with different framerates, or is that data just used directly?

Handgun/shotgun/etc reloading seems fine at both 60 & 30FPS, did they update the FCV to work at 60 for those, or is there just code somewhere to make it work at 60? (maybe the 120FPS-without-slowdown hack from #5 can help give some insight here)
(also if the FCV data for those was updated to 60, does that mean playing at 30FPS makes it run fast or something?)

If it is just FCV-data related I would have thought someone would have made a mod to fix it by now, which makes me think it's probably something code related too, that's just a wild guess though.

Maybe comparing code for Handgun/shotgun against GC could reveal something.

Keith94:
Site note: Some (or all) other issues at 60 FPS I have documented [here](https://www.pcgamingwiki.com/wiki/Resident_Evil_4_Ultimate_HD_Edition#Issues_when_running_the_game_at_60_FPS)

nipkownix:
Yeah, this is a big one.

I don't think the problems is just the FCV anim, for the same reason as you.

There are ways to import FCV anims into something like Blender nowadays, so I'll check them in comparison with other guns. 
But it is probably something related to code, though.

Version 1.0 of the game had some other issues like this. The gates in the first el gigante fight would move erratically when closing, and the gate you explode using a cannon when you first get to the castle had a _very_ odd anim. They have fixed this in ver 1.0.6, but I'm not sure if they fixed those using code or not. If they did use code, it could help us track something down by comparing what they changed.

nipkownix:
So.. I think I've found something about this:

https://user-images.githubusercontent.com/33870163/144119575-9dacc119-4b48-4649-97a2-92e6b15a25de.mp4

(I intentionally slowed down my game to make the problem more apparent)

This function is called `HermiteInterpolation`. Makes sense, judging by the name. 
Haven't investigated further yet.

emoose:
Changing one of the args (flags?) passed from `cObjSniper::moveReload` to `MotionSetCore` has an interesting effect:

[...]

PR #25, opened 2021-12-03
Add `FixSlowdown` to `FRAME RATE` settings, support for above 60FPS

emoose:
Made the fix from https://github.com/nipkownix/re4_tweaks/issues/23#issuecomment-985011951 use signatures, seems to work fine across 1.0.6/1.0.6 debug/1.1.0  😃 

I noticed one thing while testing, seems `maxFrameTime` var shouldn't be kept at 31, looks like it depends on the display / internal framerate (not sure which yet), so for now it's set to `(variableFramerate / 2) + 1`

This means going below half of the variableFramerate setting will still result in slowdown though... hopefully can find some way to detect if user is in the new-game event & only apply this maxFrameTime stuff in that case (or detect if user is in gameplay & make it disable this instead), so we can get rid of all slowdown during gameplay.

I've only added INI setting for it, wasn't sure about adding to UI since I don't think this can be reversed that easily :/

This should probably be tested out a lot before merging, have only really tried out the beginning new-game section with it myself... will probably try a 100Hz run over the weekend, if anyone has a >60Hz monitor please give it a try!

### DIFF
--- a/dllmain/60fpsFixes.cpp
+++ b/dllmain/60fpsFixes.cpp
@@ -6,6 +6,85 @@
 
 static uint32_t* ptrGameFrameRate;
 
+extern uint32_t* ptrpG; // HandleLimits.cpp
+
+int* g_ConfigVariableFramerate;
+LONGLONG* g_PrevFrameTime;
+LARGE_INTEGER* g_PerformanceFrequency;
+
+void __cdecl WinMain_DeltaTime_Hook()
+{
+	uint8_t* pG = *(uint8_t**)(ptrpG);
+
+	//void* EvtMgr = (void*)0xC03C48;
+	//uint8_t byte_C03C64 = *(uint8_t*)0xC03C64;
+
+	DWORD threadPrevAffinity = SetThreadAffinityMask(GetCurrentThread(), 0);
+
+	LARGE_INTEGER perfCount;
+	QueryPerformanceCounter(&perfCount);
+
+	int variableFramerate = *g_ConfigVariableFramerate;
+
+	double configFramerateTime = 0;
+	if (variableFramerate)
+		configFramerateTime = double(1.0) / double(variableFramerate);
+
+	// Limit game FPS to variableFramerate
+	double deltaTime;
+	do
+	{
+		QueryPerformanceCounter(&perfCount);
+		deltaTime = (double(perfCount.QuadPart) - double(*g_PrevFrameTime)) / double(g_PerformanceFrequency->QuadPart);
+		if (configFramerateTime > deltaTime)
+		{
+			Sleep(0);
+		}
+	} while (configFramerateTime > deltaTime);
+
+	// RE4 has a bug with the new-game option, when deltaTime is set too high (which happens when files are loading)
+	// seems it aborts the load for some reason, so here we'll check if deltaTime is above the lowest working FPS (seems to be variableFramerate / 2?)
+	// and if so then tell game it was variableFramerate / 2 instead
+	// (this means framerate dips below that will cause slowdown though, but those should be rare... TODO: find out what's causing the load to abort & adjust that instead)
+	// The cause of this seems to be related to R120Event, which uses cSofdec::Initialize - likely something movie related is breaking
+	// TODO: think this should be using displayFramerate / 2 instead of variableFramerate / 2, as in the actual display framerate, not sure how we could get that though
+
+	double maxFrameTime = double(1) / double((variableFramerate / 2) + 1); // seems to cancel load if it was at variableFramerate/2, but adding 1 lets it work
+
+	if (variableFramerate != 30 && deltaTime > maxFrameTime
+		// TODO: seems this is trying to check for a problematic event, maybe trying to check for the new-game load screen?
+		// doesn't seem to work for new-game bug though sadly, so commented out for now so this can be applied anywhere
+		// (would be great to find a way to limit this so it's event-only, so that going under (variableFramerate / 2) would also have slowdown fixed)
+		// && EventMgr__IsAliveEvt(EvtMgr, 0, byte_C03C64, 0, 0)
+		)
+	{
+		deltaTime = maxFrameTime;
+	}
+	// We want to limit update-FPS to variableFramerate, so change deltaTime to configFramerateTime if deltaTime is lower
+	else if (deltaTime < configFramerateTime)
+	{
+		deltaTime = configFramerateTime;
+	}
+
+	*(float*)(pG + 0x70) = float(deltaTime * 30);
+
+	SetThreadAffinityMask(GetCurrentThread(), threadPrevAffinity);
+
+	*g_PrevFrameTime = perfCount.QuadPart;
+}
+
+void FPSSlowdownFix_GetPointers()
+{
+	auto pattern = hook::pattern("A3 ? ? ? ? 8B 46 ? D9 1D ? ? ? ? DB 46 3C 89 0D ? ? ? ?");
+	g_ConfigVariableFramerate = (int*)*pattern.count(1).get(0).get<uint32_t>(0x13);
+
+	pattern = hook::pattern("8B 8D ? ? ? ? 8B 95 ? ? ? ? 89 0D ? ? ? ? 89 15 ? ? ? ? E8 ? ? ? ? 8D 85");
+	g_PrevFrameTime = (LONGLONG*)*pattern.count(1).get(0).get<uint32_t>(0xE);
+
+	pattern = hook::pattern("FF 15 ? ? ? ? 50 FF 15 ? ? ? ? 68 ? ? ? ? 8B F0 FF D7 68 ? ? ? ? FF D3");
+	g_PerformanceFrequency = (LARGE_INTEGER*)*pattern.count(1).get(0).get<uint32_t>(0xE);
+}
+
 void Init_60fpsFixes()
 {
 	auto pattern = hook::pattern("89 0D ? ? ? ? 0F 95 ? 88 15 ? ? ? ? D9 1D ? ? ? ? A3 ? ? ? ? DB 46 ? D9 1D ? ? ? ? 8B 4E ? 89 0D ? ? ? ? 8B 4D ? 5E");
@@ -44,4 +123,26 @@ void Init_60fpsFixes()
 				_asm {fmul vanillaMulti}
 		}
 	}; injector::MakeInline<AmmoBoxSpeed>(pattern.count(2).get(1).get<uint32_t>(0), pattern.count(2).get(1).get<uint32_t>(6));
+
+
+	// Fix for slowdown when FPS drops, and allow using framerates above 60
+	// (reimplemented part of WinMain)
+	FPSSlowdownFix_GetPointers();
+
+	if (cfg.bFixFPSSlowdown)
+	{
+		pattern = hook::pattern("89 15 ? ? ? ? A3 ? ? ? ? 6A 00 FF 15 ? ? ? ? 50");
+		uint8_t* WinMain_DeltaTime_start = pattern.count(1).get(0).get<uint8_t>(0xB);
+		injector::MakeCALL(WinMain_DeltaTime_start, WinMain_DeltaTime_Hook, true);
+
+		// fix ESP, not sure why this is needed...
+		uint8_t dec_esp_4[] = { 0x83, 0xC4, 0xFC };
+
+		injector::WriteMemoryRaw(WinMain_DeltaTime_start + 5, dec_esp_4, 3, true);
+
+		// JMP over stuff that we reimplemented...
+		pattern = hook::pattern("E8 ? ? ? ? 8D 85 ? ? ? ? 50 FF D3 DF AD ? ? ? ? 8D 8D ? ? ? ? 51 DC 35 ? ? ? ? DC 25 ? ? ? ? DD 1D ? ? ? ?");
+		uint8_t* WinMain_DeltaTime_end = pattern.count(1).get(0).get<uint8_t>(0);
+		injector::MakeJMP(WinMain_DeltaTime_start + 5 + 3, WinMain_DeltaTime_end, true);
+	}
 }
--- a/dllmain/Settings.cpp
+++ b/dllmain/Settings.cpp
@@ -172,6 +172,7 @@ void Settings::ReadSettings()
 	cfg.bRaiseVertexAlloc = iniReader.ReadBoolean("MEMORY", "RaiseVertexAlloc", true);
 	cfg.bRaiseInventoryAlloc = iniReader.ReadBoolean("MEMORY", "RaiseInventoryAlloc", true);
 	cfg.bUseMemcpy = iniReader.ReadBoolean("MEMORY", "UseMemcpy", true);
+    cfg.bFixFPSSlowdown = iniReader.ReadBoolean("FRAME RATE", "FixSlowdown", false);
 	cfg.bIgnoreFPSWarning = iniReader.ReadBoolean("FRAME RATE", "IgnoreFPSWarning", false);
 }
 
--- a/dllmain/Settings.h
+++ b/dllmain/Settings.h
@@ -41,6 +41,7 @@ struct Settings
 	bool bRaiseVertexAlloc;
 	bool bRaiseInventoryAlloc;
 	bool bUseMemcpy;
+	bool bFixFPSSlowdown;
 	bool bIgnoreFPSWarning;
 
 	bool HasUnsavedChanges;
--- a/settings/settings_string.h
+++ b/settings/settings_string.h
@@ -101,6 +101,11 @@ RaiseInventoryAlloc = true
 UseMemcpy = true
 
 [FRAME RATE]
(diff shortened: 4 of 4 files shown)


Keith94:
I played at 144 FPS up until the Hunnigan cutscene. Didn't encounter any issues so far.

nipkownix:
Nice! I'll test the stuff that triggers the usual FPS-related bugs, like this one I recorded on release:

https://www.youtube.com/watch?v=PZEGNNjldqE

~geez, has it been this long already?!~

emoose:
Huh I've never seen that grenade problem before, that was with 60FPS? Wonder what could cause that.

>I have tried the new build along with resolution 1920 1080 144; however, game is still running at 60 Hz (in fullscreen). Can't really get Special K to work either for some reason...

Ah yeah same happens here, I only really tested this in windowed mode, but looks like fullscreen always uses 60Hz mode for some reason, will need to try looking into what handles that.

emoose:
Added some hooks that should allow non-60Hz fullscreen modes hopefully, like mentioned above I haven't been able to test it properly myself though, but think it should work.

Build+INI here: [winmm_fps-slowdown-fix-v2.zip](https://github.com/nipkownix/re4_tweaks/files/7652145/winmm_fps-slowdown-fix-v2.zip)

Make sure the `resolution` line in config.ini is set with the refresh rate you want to use, and hopefully with that the game should be able to use that rate in fullscreen.

nipkownix:
> Huh I've never seen that grenade problem before, that was with 60FPS? Wonder what could cause that.

The first video (from 2014) was at 59 fps. Yeah, even one frame less than 60 caused issues.
The other video was at 144 fps.

> Added some hooks that should allow non-60Hz fullscreen modes hopefully, like mentioned above I haven't been able to test it properly myself though, but think it should work.

Oh, nice! This should fix the black screen issue people have when their monitors don't have 60 hz as an option.

Keith94:
Great finds thanks!! I'll try it out when I get home and get back to you.

emoose:
> Edit: latest build is crashing inside gameMainLoop, at this instruction:
> 
> image

That's with the build I just posted? What game ver is that? I did test against 1.1.0/1.0.6 before pushing, not sure why it'd be changing gameMainLoop.
Does disabling `FixDisplayMode` get rid of it? Maybe the sig is matching some other piece of code somewhere.

nipkownix:
Hmm, tried disabling both options, but the game still crashes. There might be something wrong with my game? I'm using .exe ver 1.1.0

emoose:
Ah looks like it freezes on quit for me too with 60FPS, looks like it's the `FixSlowdown` fix that causes that, hmm

Does the inventory issue go away if you disable both FixSlowdown & FixDisplayMode?

DustyCrumpet:
Here's a few issues:

1) Audio issues with footsteps. 

https://user-images.githubusercontent.com/58010930/144673171-754ba0d8-b55f-4bb6-b8f0-c5ce1ba84d03.mp4

2) Backwards turning is too fast.

https://user-images.githubusercontent.com/58010930/144673732-99f78203-eee9-4193-9b2c-92cbcc225ab4.mp4

3) Striker spews out shells after firing / TMP reload sound repeats.

https://user-images.githubusercontent.com/58010930/144673853-1732fb82-90b2-43c4-a18e-b07f28042d7a.mp4

4) Backing out of the inventory can cause the game to remain paused (Health bars still animate).

https://user-images.githubusercontent.com/58010930/144674040-a2da0924-67c4-43ad-983c-3ff1fb12fb08.mp4

5) Water and Leon's headturning may be too fast.

https://user-images.githubusercontent.com/58010930/144674089-253f3348-9c3d-4d5f-b608-4651e631e43a.mp4

60FPS footage was recorded with FixSlowdown = false and variableframerate 60.

emoose:
> Here's a few issues:

Thanks for giving it a try, I'd guess some of these like backwards turning/head turn/water are probably also wrong speed at 60FPS too, probably need to find a way to adjust those to match with 30FPS speed, like how nipkownix's 60FPS item falling fixes   were done.

That striker bug is crazy lol, haven't seen anything like that before, guess that's something animation related.

nipkownix:
>Does the inventory issue go away if you disable both FixSlowdown & FixDisplayMode?

It doesn't, actually. 

This makes no sense.

Release/Debug build from master: no inventory issue.
Your build with both new options disabled: inventory issue shows.

However, people on nexus were having this exact same issue with the latest release, while it works fine for me.

~Cursed bug~

emoose:
> > Does the inventory issue go away if you disable both FixSlowdown & FixDisplayMode?
> 
> It doesn't, actually.
> 
> This makes no sense.
> 
> Release/Debug build from master: no inventory issue. Your build with both new options disabled: inventory issue shows.
> 
> However, people on nexus were having this exact same issue with the latest release, while it works fine for me.
> 
> ~Cursed bug~

Damn... guessing it's probably some weird memory issue, maybe something gets corrupt under certain conditions, I hate bugs like that >.<

For me the inventory seems fine with this build, same with your one on nexus too :/

I'll try looking over the stuff we patched to redirect the buffer soon, maybe something was left out somewhere.

(wonder if maybe g_MemPool_SubScreen/g_MemPool_SubScreen2 initial values could be affecting anything, maybe those should be memset to 0 first...)

nipkownix:
> E: btw just to be sure, you're using the same resolution/fullscreen mode/etc between tests right? Was thinking maybe screen res changes how much of those buffer is filled, not sure about that though.

Yep, same settings and starting a new game on every test.

Keith94:
>Build+INI here: winmm_fps-slowdown-fix-v2.zip

Can confirm that `FixDisplayMode = true` fixes the refresh rate issue, so the game defaults to whatever I set in Windows. :)

However the game is unstable in this build. Alt+Tabbing and opening the video settings in game will result in crash.

emoose:
Looks like the striker shell issue is something to do with the call to `MotionCheckCrossFrame` inside `cObjStriker::moveFire`, some reason at 144Hz that's always returning true, which makes it call into `cObjStriker::setCartridge` to spawn shells everywhere.

Not sure why `MotionCheckCrossFrame` is broken, probably some float it's checking is off. I've been trying to map out the MOTION_INFO structs etc, mostly have the field types mapped out but haven't named much yet. IDB is here if anyone wants to have a look at it: [pc-1.1.0_enum - Copy.7z.zip](https://github.com/nipkownix/re4_tweaks/files/7653416/pc-1.1.0_enum.-.Copy.7z.zip)

nipkownix:
>E2: oh wait wtf, the shell issue happens for me even with variableframerate 60 in config.ini, probably not framerate related then, guess it's something to do with the deltaTime calculation or something :/

Yep, I was about to post that.
Pretty wild bug.. looks hilarious! I even got to change the direction of the shells and everything.

emoose:
Hmm, might have an idea what's happening

```C
      v4 = (flags & 4) == 0;
      animNumFrames_ = a1->numFrames_BC;
      newFrameNum = *((float *)pG + 0x1C) * a1->speed_C0 + a1->field_B8;
      newFrameNum_ = newFrameNum;
      a1->field_B8 = newFrameNum;
      if ( v4 )
      {
        animNumFrames__ = (double)animNumFrames_;
        if ( animNumFrames__ <= newFrameNum_ )
        {
          v9 = animNumFrames__ - *((float *)pG + 0x1C);
          a1->field_42 = 4;
          a1->field_B8 = v9;
        }
      }
      else
      {
        v7 = (double)animNumFrames_;
        if ( v7 <= newFrameNum_ )
        {
          a1->field_42 = 1;
          a1->field_B8 = newFrameNum_ - v7;
        }
      }
    }
    
```
In that code it seems to be using the deltaTime (`*((float *)pG + 0x1C)`) multiplied by anims speed to find the new frame num, then it checks that frame num against the total number of frames to make sure it can show it, if new frame num is larger than frame count, it tries removing value of deltaTime from it (I guess this would keep removing deltaTime from it each frame until new frame num is valid)
Problem is with our `FixSlowdown` setting, deltaTime isn't fixed any more, and probably varies a lot each frame, so it'd probably end up overshooting the frame num every time.

Wonder if changing that so it just uses animNumFrames__  value if it's above it might help at all, wouldn't be surprised if that ends up breaking other stuff though.

emoose:
Here's a build of that commit:
[winmm_fps-slowdown-fix-v3.zip](https://github.com/nipkownix/re4_tweaks/files/7653875/winmm_fps-slowdown-fix-v3.zip)

Only difference I've noticed is the striker shell bug is fixed, looks like grenades are still broken, and game still won't quit properly :/

emoose:
Alright seem to have something working, not sure if it might be breaking something else though, but seems to fix grenades at least.

Build+INI: [winmm_fps-slowdown-fix-v4.zip](https://github.com/nipkownix/re4_tweaks/files/7654032/winmm_fps-slowdown-fix-v4.zip)

nipkownix:
Nice! Your fixes are working nicely here!
The only obvious problems I've noticed are random freezes. Sometimes it freezes when loading a save, for some reason.
Also, I've noticed the aim speed was inconsistent. Apparently it got slower with high frame rates. I've pushed a fix that should make everything more consistent.

DustyCrumpet:
> Here's a build of that commit: [winmm_fps-slowdown-fix-v3.zip](https://github.com/nipkownix/re4_tweaks/files/7653875/winmm_fps-slowdown-fix-v3.zip)
> 
> Only difference I've noticed is the striker shell bug is fixed, looks like grenades are still broken, and game still won't quit properly :/ I haven't tested it that much yet though, maybe it'd help with some other issues too.

This build fixes the TMP reload sound too, but breaks several death animations (even at 60fps with FixSlowdown = true):

https://user-images.githubusercontent.com/58010930/144715715-240398e7-0297-4d51-ad7b-0e86c2b4d34c.mp4

Also another bug that happens with the original build (second jump on the church bridge):

https://user-images.githubusercontent.com/58010930/144715733-0ce8f4bc-d2b7-4795-9d65-fe57a478c624.mp4

nipkownix:
About `FixDisplayMode`, I'd suggest maybe getting the refresh rate from Window's current resolution instead of reading it from config.ini. This would solve people's black screen issues without having to change anything inside config.ini.

A custom refresh rate could then be set from winmm.ini, which is easier to find and allows evrything to be configured from one place, + we can change it using the ImGui menu, I guess.

SupahDrivah1234:
Hi, I've been having trouble trying to get the Steam version of Resident Evil 4 running on my PC for the past year or so since my laptop (a Lenovo Legion 7i) has no selectable refresh rate other than 144Hz and CRU isn't working for me. The game would only open in a small black window with audible sound but no video.

This fix managed to at least fix the video, for which I'm grateful, but I hadn't walked two feet before trying to check settings in the options menu and it crashes to desktop. 

As far as I can tell, it's only the primary pause/options menu. Is there a way to fix this?

nipkownix:
Made the changes to `FixDisplayMode` I mentioned earlier, and implemented it in the current master. I wanted to get this fix in re4_tweaks before the HD project releases, since a large number of people with non-60Hz displays will likely run into this issue when trying RE4 for the first time with the HD project.

@SupahDrivah1234, I believe you're the same person I replied to on Steam, but here's a build with just the black screen fix and nothing else from this PR, since the other stuff is probably what is crashing your game:
[Bin32.zip](https://github.com/nipkownix/re4_tweaks/files/7891321/Bin32.zip)

Mister-Curious:
about that grenade suspension problem; was this ever resolved? I recently noticed this too on a room I was working on, but managed to fix it by editing the EAT file. Maybe that issue you see is simply some bad EAT in the room you were in  (looks like r109 if I am not mistaken). If the issue is still present, I can edit the r109 and send it over for testing.

SupahDrivah1234:
> Made the changes to `FixDisplayMode` I mentioned earlier, and implemented it in the current master. I wanted to get this fix in re4_tweaks before the HD project releases, since a large number of people with non-60Hz displays will likely run into this issue when trying RE4 for the first time with the HD project.
> 
> @SupahDrivah1234, I believe you're the same person I replied to on Steam, but here's a build with just the black screen fix and nothing else from this PR, since the other stuff is probably what is crashing your game: [Bin32.zip](https://github.com/nipkownix/re4_tweaks/files/7891321/Bin32.zip)

I think you did reply to me on Steam, yes. 

Unfortunately, even with a clean install of RE4 this new build didn't work. I'm facing the same issue of being unable to open the options menu without crashing to desktop.

nipkownix:
Hmm, this new test build has some logging that might help us understand the issue a little better:
[winmm.zip](https://github.com/nipkownix/re4_tweaks/files/7897985/winmm.zip)

To the get log, create a folder named "CrashDumps" in re4's bin32 folder, near the .exe file.
Then, have the game crash as usual, and two files should be generated inside that new folder.
Upload them here so we can have a look.

emoose:
>Maybe the game is making a list of 60hz resolutions -only-? And we end up with an empty list if no 60hz mode is present?

That reminded me of this comment in one of the commits here: https://github.com/nipkownix/re4_tweaks/pull/25/commits/28c27015797f575b2b45a4e36b94e43c49afa38e#diff-59bafca88011950dfe128dac79f2499e832dde20dc1ee96cdf788919734f3961R170

Could be that's related, I'm not sure if your FixDisplayMode touches the func hooked there, might be worth looking into.

nipkownix:
Hmm, my FixDisplayMode is the exact same as yours, the only difference being that I get the refresh rate from Windows instead of getting it from the ini file.

emoose:
I think that fix only hooks the EnumAdapterModes stuff in 0x9391E0, but looks like the 0x953480 func that calls EnumAdapterModes also has some weird 60 checks inside it, maybe that's what makes the res list.

nipkownix:
Oh, there's another func calling EnumAdapterModes? I'll check that out.

nipkownix:
@SupahDrivah1234, let me know if this build resolves your issue:
[winmm.zip](https://github.com/nipkownix/re4_tweaks/files/7900104/winmm.zip)

Mister-Curious:
@nipkownix Glad to hear it was resolved. EAT files handle projectile collision data and has the same structure as SAT which handles player and enemy collision data. With no EAT there is no ground for the grenade to bounce off.

SupahDrivah1234:
> @SupahDrivah1234, let me know if this build resolves your issue: [winmm.zip](https://github.com/nipkownix/re4_tweaks/files/7900104/winmm.zip)

That worked! thank you so much for your help.

emoose:
BTW the PCGW page for RE4 currently links to the old `winmm_fps-slowdown-fix-v2.zip` build in this PR as part of fixes for non-60Hz screens: https://www.pcgamingwiki.com/wiki/Resident_Evil_4_Ultimate_HD_Edition#No_video_on_monitors_with_high_refresh_rates_.2F_Game_runs_at_incorrect_refresh_rate

Maybe it's worth making a new proper release, if you feel the new changes after 1.6 are stable enough.

nipkownix:
Ah, thanks for the heads up. I'll probably make a new release today, then. I just want to address #40 before making the release.

emoose:
Hmm, so the main animation problems seemed to be caused by delta-time varying between different frames... I wonder if locking it to 90/120FPS frametime would work better, wouldn't have the benefit of slowdown fix, but it'd give benefit of increased framerate at least, I guess it'd probably help show what things need framerate fixes too (wind/water/etc).

Maybe I'll make a new PR soon with something like that, since this PR draft is mostly to do with slowdown.

nipkownix:
>Maybe I'll make a new PR soon with something like that, since this PR draft is mostly to do with slowdown.

Nice. Sounds like a good idea to me.

emoose:
Huh seems it's easier than I thought, just need to overwrite the double it uses for limiting framerate from 0.166666... to 1/framerate, so 0.011111... for 90FPS, (double is at 0xB1D180 in 1.1.0), then make sure your config.ini variableframerate is set to 90, and with those game looks like it runs at 90 fine - would probably work with 120 fine as well, but my monitor doesn't support that so can't test it myself - not sure if 144 would work that well since it's not a multiple of 30.

(btw, seems re4_tweaks is ignoring the `IgnoreFPSWarning` ini setting now for some reason? always shows the messagebox about framerate being changed, had to edit it with CE instead...)

nipkownix:
>(btw, seems re4_tweaks is ignoring the IgnoreFPSWarning ini setting now for some reason? always shows the messagebox about framerate being changed, had to edit it with CE instead...)
E: building from src seems to have fixed that, weird.

Could have been because I moved IgnoreFPSWarning to another section in the .ini file?

emoose:
> Could have been because I moved IgnoreFPSWarning to another section in the .ini file?

Ah that might be it, looks like my INI had it in the new section tho, so guess I was probably using a build from before 1.7.2 somehow (maybe built from an older commit at some point)

In case anyones wondering, the builds I posted are from the latest commit, just added a line of code to patch the frametime in.

Keith94:
Not sure if it's a known issue, but I tried `re4_tweaks1.7.2_144fps.zip` and the inputs in the menu are twice as fast, so it often skips two menu options with one press.

emoose:
Posted a new PR with improved framelimiter fix here: https://github.com/nipkownix/re4_tweaks/pull/50

If anyone wants to test greater-than-60 FPS please let us know of any issues there!

PR #26, opened 2021-12-03
Minor `RaiseInventoryAlloc` improvements

emoose:
This'll now hook the `mem_alloc` call inside `MessageControl::stageInit` to change the `pG + 0x3C` pointer, instead of hooking the whole stageInit func and changing it afterwards, this should now let the `MessageControl::loadStageFont` call inside stageInit act on our new buffer instead, maybe will help with inventory issue, not sure.

Also memsets to 0 & aligns the SubScreen buffers to 4096 bytes, and in debug builds adds a SafetyBuf between them, which will cause an exception if anything writes to it - would probably be good to have this in release too, but 0x4000000 bytes is a lot to waste on it...

Sig seems to work fine across 1.1.0/1.0.6/debug.

### DIFF
--- a/dllmain/HandleLimits.cpp
+++ b/dllmain/HandleLimits.cpp
@@ -7,7 +7,7 @@
 uint32_t* ptrpG;
 
 static uint32_t* ptrpzzl_size;
-static uint32_t* ptrstageInit;
+static uint32_t* ptrstageInit_mem_alloc_call;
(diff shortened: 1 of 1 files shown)


nipkownix:
Nice! I'll see if I can get the inventory issue to show up again, and also send a test build to the people in Nexus. Thanks!

PR #30, opened 2022-01-06
Add `DebugMenuKeyCombination` setting, allows custom tool menu key combo

emoose:
New default for tool menu is now `CTRL+F3`, as not all keyboards have HOME+END, and F1/F2 are used by Imgui things - the combo string will be parsed into a std::vector, which the tool-menu hook iterates over to check if all keys are pressed.

(the combo uses multiple keys to reduce chance of it being pressed by accident, don't want people to get stuck in the tool-menu mode...)

This also changes Setting.cpp so that `key_map` is defined statically (afaik previously it looked like key_map was inited each time KeyMap got called), and changed the `char* type` KeyMap param to a bool instead to get rid of a string compare.

Also added some key name variants to key_map to improve chance of users combo string getting parsed, there were a couple keys that I added to key_map but don't know the DIK of tho, what does DIK mean anyway? DirectInput key or something?

E: build here if anyone wants to test it:
[winmm.zip](https://github.com/nipkownix/re4_tweaks/files/7820502/winmm.zip)

E2: DIKless keycodes:
- RENTER (numpad enter key, i think)
- CONTROL/LCONTROL/CTRL/LCTRL
- ALT/LALT
- RCONTROL/RCTRL
- ALTGR/RALT (AltGr is UK keyboard name for right alt key)
- ;
- :
- /
- ?
- '
- @
- [
- {
- \
- |
- ]
- }
- \#
- "
- `
 
The symbols there are mostly VK_OEM_* keys, I only really know the OEM symbols for my UK keyboard, which are what I added there, if someone has US keyboard it'd be good if you can try finding what symbols match up to VK_OEM_1/OEM_2/etc.
Not sure if those symbols have DIKs for us to add in there neither, the tool menu stuff doesn't need any DIK though so not a big deal there, maybe worth adding for the hooks that do use it though.

### DIFF
--- a/dllmain/Settings.cpp
+++ b/dllmain/Settings.cpp
@@ -5,126 +5,193 @@
 #include "Settings.h"
 #include "settings_string.h"
 
+// tool_menu.cpp
+bool ParseToolMenuKeyCombo(std::string_view in_combo);
+
 Settings cfg;
 
-int Settings::KeyMap(char* key, const char* type)
+struct key_type {
+	int vk;
+	int dik;
+};
+
+std::unordered_map<std::string, key_type> key_map
 {
-	struct key_type {
-		int vk;
-		int dik;
-	};
+	{ "ESCAPE", { 0x1B, 0x01 } },    // Esc
+	{ "0", { 0x30, 0x0B } },    // 0
+	{ "1", { 0x31, 0x02 } },    // 1
+	{ "2", { 0x32, 0x03 } },    // 2
+	{ "3", { 0x33, 0x04 } },    // 3
+	{ "4", { 0x34, 0x05 } },    // 4
+	{ "5", { 0x35, 0x06 } },    // 5
+	{ "6", { 0x36, 0x07 } },    // 6
+	{ "7", { 0x37, 0x08 } },    // 7
+	{ "8", { 0x38, 0x09 } },    // 8
+	{ "9", { 0x39, 0x0A } },    // 9
+	{ "A", { 0x41, 0x1E } },    // A
+	{ "B", { 0x42, 0x30 } },    // B
+	{ "C", { 0x43, 0x2E } },    // C
+	{ "D", { 0x44, 0x20 } },    // D
+	{ "E", { 0x45, 0x12 } },    // E
+	{ "F", { 0x46, 0x21 } },    // F
+	{ "G", { 0x47, 0x22 } },    // G
+	{ "H", { 0x48, 0x23 } },    // H
+	{ "I", { 0x49, 0x17 } },    // I
+	{ "J", { 0x4A, 0x24 } },    // J
+	{ "K", { 0x4B, 0x25 } },    // K
+	{ "L", { 0x4C, 0x26 } },    // L
+	{ "M", { 0x4D, 0x32 } },    // M
+	{ "N", { 0x4E, 0x31 } },    // N
+	{ "O", { 0x4F, 0x18 } },    // O
+	{ "P", { 0x50, 0x19 } },    // P
+	{ "Q", { 0x51, 0x10 } },    // Q
+	{ "R", { 0x52, 0x13 } },    // R
+	{ "S", { 0x53, 0x1F } },    // S
+	{ "T", { 0x54, 0x14 } },    // T
+	{ "U", { 0x55, 0x16 } },    // U
+	{ "V", { 0x56, 0x2F } },    // V
+	{ "W", { 0x57, 0x11 } },    // W
+	{ "X", { 0x58, 0x2D } },    // X
+	{ "Y", { 0x59, 0x15 } },    // Y
+	{ "Z", { 0x5A, 0x2C } },    // Z
+	{ "BACK", { 0x08, 0x0E } },    // Backspace
+	{ "ADD", { 0x6B, 0x4E } },    // Numpad +
+	{ "DECIMAL", { 0x6E, 0x53 } },    // Numpad .
+	{ "DIVIDE", { 0x6F, 0xB5 } },    // Numpad /
+	{ "MULTIPLY", { 0x6A, 0x37 } },    // Numpad *
+	{ "NUMPAD0", { 0x60, 0x52 } },    // Numpad 0
+	{ "NUMPAD1", { 0x61, 0x4F } },    // Numpad 1
+	{ "NUMPAD2", { 0x62, 0x50 } },    // Numpad 2
+	{ "NUMPAD3", { 0x63, 0x51 } },    // Numpad 3
+	{ "NUMPAD4", { 0x64, 0x4B } },    // Numpad 4
+	{ "NUMPAD5", { 0x65, 0x4C } },    // Numpad 5
+	{ "NUMPAD6", { 0x66, 0x4D } },    // Numpad 6
+	{ "NUMPAD7", { 0x67, 0x47 } },    // Numpad 7
+	{ "NUMPAD8", { 0x68, 0x48 } },    // Numpad 8
+	{ "NUMPAD9", { 0x69, 0x49 } },    // Numpad 9
+	{ "OEM_COMMA", { 0xBC, 0x33 } },    // OEM_COMMA (< ,)
+	{ "OEM_MINUS", { 0xBD, 0x0C } },    // OEM_MINUS (_ -)
+	{ "OEM_PERIOD", { 0xBE, 0x34 } },    // OEM_PERIOD (> .)
+	{ "OEM_PLUS", { 0xBB, 0x0D } },    // OEM_PLUS (+ =)
+	{ "<", { 0xBC, 0x33 } },    // OEM_COMMA (< ,)
+	{ ",", { 0xBC, 0x33 } },    // OEM_COMMA (< ,)
+	{ "_", { 0xBD, 0x0C } },    // OEM_MINUS (_ -)
+	{ "-", { 0xBD, 0x0C } },    // OEM_MINUS (_ -)
+	{ ">", { 0xBE, 0x34 } },    // OEM_PERIOD (> .)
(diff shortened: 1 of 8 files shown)


emoose:
Should also fix the debug menu selection on 1.0.6 now too, haven't tested on 1.1.0 & 1.0.6debug yet though, but hopefully those should still work as they did before.

nipkownix:
>Also added some key name variants to key_map to improve chance of users combo string getting parsed, there were a couple keys that I added to key_map but don't know the DIK of tho, what does DIK mean anyway? DirectInput key or something?

I think it means DirectInput Keyboard, not sure.
https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ee418641(v=vs.85)

As far as I know, all of this was deprecated by Microsoft even before this port released, but RE4 used it anyways, so... meh.
All the bindings in the game refer to those.

>E2: DIKless keycodes:
>
>    RENTER (numpad enter key, i think)
>    CONTROL/LCONTROL/CTRL/LCTRL
>    ALT/LALT
>    RCONTROL/RCTRL
>    ALTGR/RALT (AltGr is UK keyboard name for right alt key)

Hmm, some of these are there thought, aren't they?
`{ "RCONTROL", { 0xA3, 0x9D } },    // Right Ctrl`

0x9D being the DIK one.

Buut, after that there's a duplicate entry(?) without it:
`{ "RCONTROL", { VK_RCONTROL, 0 } },  // Right Control`

Mister-Curious:
Thanks again for looking into this. Just porting back here on this build.  The Ctrl + F3 works well (even with AHK sending the keystrokes), so thanks for that. 

The issue of Debug menu speed using a controller was not resolved for me in this build. I tested it with and without DS4Windows running, but the result is the same. The Debug menu is still accepting input at an accelerated rate making navigation with a controller impossible. I will revert to my AHK binding of the Dpad to keyboard arrows for the time being. 

Also I am having a REALLY hard time reading the console and F2 menus as the text size is incredibly small on my 1080p TV. Maybe its my old dinosaur eyes, but it would be helpful for aging people, and those with visual impairments to have the option to set the text size. 

Cheers!

Mister-Curious:
Sorry I'm back again. Another issue with this build is that is changes the controller sensitivity. It's really noticeable as the controller is much to sensitive. Not sure what's going with this, but though it must be related to the most recent work. 

Thanks!

nipkownix:
The debug window will be disabled on release builds, so no need to worry about it too much.

About the font size: I could add some buttons to increase text sizes. I'll look into that tomorrow.

The controller sensitivity isn't a bug, though. It is intentional. It is part of the "60fpsfixes" option. It makes the aiming speed consistent with how it feels when playing at 30 fps, which was the intended speed to begin which. Let me know if the speed is actually faster than when playing at vanilla 30 fps, though.

Mister-Curious:
Thanks again for a speedy reply. Good to know about the overlay thing. 

I thought the controller response input in main game play was a bug because it wasn't present in the other build I was using, only in the one I got from #30. I have never played with 30 fps so I couldn't really tell you about that. So let me ask, if I don't want my controller sensitivity changing while using the DLL, would I simply disable the FPS fix?

Also, lastly wondering... Were you able to take another look at the Debug controller input speed ? It sounded like you fixed it on your end, but it still is wonky for me. 

Thanks again. Let me know when you have some time for a chat on Discord. Ive sent a couple of offline messages just to see if you were around. 

Cheers

emoose:
Did you try the build linked at https://github.com/nipkownix/re4_tweaks/pull/30#issuecomment-1006381783? That one should have debug speed fix included, the build in first post only has the custom key bindings.

Not sure how I missed the duplicate key names, I'll fix that up & try adding in missing DIKs soon.

Mister-Curious:
Oh I guess I missed that link in the second post. The fix works great thank you. I look forward to speaking with you at some point. 

Cheers

Mister-Curious:
oh another thing. As I mentioned, with these latest side-builds my controller becomes quite sensitive, so I am wondering if there is an option to preserve the sensitivity? I have set 60fpsFixes = false but the controller remains too sensitive.

emoose:
Removed dupes from key_map & added in DIK values for the new entries, except for some of the localised VK_OEM_* ones, the 5 that still have DIK = 0 I think have correct VK set for them, but DIK seems to swap between different layouts, not sure if there's any easy way to set up the DIK values for them based on users KB layout or something like that, could probably just remove them instead really.

Mister-Curious:
> @Mister-Curious do you know what the last build of re4_tweaks you used that didn't have changed sensitivity was? Was that the 0.6 release or did you build from a commit or something? Maybe can track down which commit changed it for you.

I was thinking the same thing, but I don't know which version it is. In the Details pane of the file itself it says 1.6.0.0. but when I compare it to the version I am using now they are different (compared in HEX editor). 

I have [uploaded](https://drive.google.com/file/d/1V5qK5q6ctjj288Cb2ggCUHyo3bDaY-cB/view?usp=sharing) the version I am using in hopes you might be able to compare it to any versions you have

Thanks again.

nipkownix:
Hmm, does this build make the issue go away?

[winmm.zip](https://github.com/nipkownix/re4_tweaks/files/7846343/winmm.zip)

If it does, can you show us whats the output in the debug window?

Mister-Curious:
> Hmm, does this build make the issue go away?
> 
> [winmm.zip](https://github.com/nipkownix/re4_tweaks/files/7846343/winmm.zip)
> 
> If it does, can you show us whats the output in the debug window?

Thanks for this. The controller sensitivity issue is still present in this version you sent.

nipkownix:
Oh, really? Weird. What about this one?
[winmm.zip](https://github.com/nipkownix/re4_tweaks/files/7850141/winmm.zip)

Mister-Curious:
THey thanks again, but still no dice. Its still overly sensitive with this last DLL you sent.

nipkownix:
I see. Then it is probably something realted to the mouse turning feature.

This one should make the issue go away:

[winmm.zip](https://github.com/nipkownix/re4_tweaks/files/7850650/winmm.zip)

The problem is that every change I made with that feature was tested using mouse and keyboard only. Meh.

Mister-Curious:
> The problem is that every change I made with that feature was tested using mouse and keyboard only. Meh.

This version is fine, no controller issues. Does that mean you can merge something from this build with the #30 build that fixes the other controller stuff?

Thanks!

nipkownix:
This build should have everything fixed, including the changes from this PR:
[winmm.zip](https://github.com/nipkownix/re4_tweaks/files/7854912/winmm.zip)

Thanks once again, emoose!

@Mister-Curious, I added some checks to the mouse turning feature so they only take effect if you're not using a controller. This probably needs some more testing though, so feel free to open another issue if something else is broken. Thanks!

Mister-Curious:
works solid for the issues we discussed. Thank you !

Mister-Curious:
weird thing though. When I use my modded binary this works fine, but with the vanilla 1.0.6 version (with 4gb patch applied) it crashes. What is going on ?

nipkownix:
Crashes on start up? Any error message? Could you upload your .exe?

Mister-Curious:
https://drive.google.com/file/d/1Y8SNLpAW_6dyRfVQj13CfhtLCsnAC7S3/view?usp=sharing

bio4 version 1.0.6 patched

Mister-Curious:
ok so I had my engineer friend check on his machine, same thing..instant crash with vanilla exe, but our modded version works fine. Whats the difference? Our exe remove's the splash screen that is used by Raz0r's DLL (many people use this DLL, including me for my mod). So we tested your latest DLL without Raz0r's dll with vanilla exe and it works fine. 

The going theory is that your DLL injecting itself during Raz0r's splash screen produces an error for the dll since it does not account for that situation resulting in an instant crash
At the top right we see something that says "Press F1 to view console", Raz0r also puts a disclaimer somewhere around there too, I'm guessing this might cause a conflict

nipkownix:
Ah, Raz0r's stuff.
His dll is trying to replace the code I use to display the config menu with his code to display his.. uh.. amazingly necessary "bsaa anti-piracy screen". This ends up creating some broken mess and the game crashes.

nipkownix:
Let me know if this build fixes it:
[winmm.zip](https://github.com/nipkownix/re4_tweaks/files/7862816/winmm.zip)

Mister-Curious:
Thank you, this resolved the issue (its loads really quickly now). I will ensure that any testing I do in the future includes the vanilla exe as well.

emoose:
@Mister-Curious btw what are you using from Raz0rs DLL in your mod, do you need one of the patches it applies or something? If it's not something too involved maybe possible to replicate it in re4_tweaks.

Mister-Curious:
This was something I wanted to discuss with you on Discord if possible. I could also email if that's better. For the time being I can outline what functions it allows me to do:

- side chain files  (has config, and directories)
- replace EVD cutscene files with movie files  (has config)
- play movies anywhere (has config)
- remove Ashely from certain rooms (has config)
- lock certain AEV events with key item (has config)
- skip chapter end screens and radio calls
- swap TMP for custom flamethrower (has config)

As you can see there are lots of things I rely on, but many of these things I could get my engineer friend to hardcode, but the configurable stuff would need a DLL with config file.

Mister-Curious:
@emoose Seeing that the list of things I am dependent on from Raz0r's DLL is long, perhaps it is too much work to try and replicate everything he is doing (although I would be much more pleased to use your DLL instead). If it is not possible to replicate all of these functions, then perhaps I can get a version of your DLL that is watered down to have at least no conflicts with the memory allocation etc. 

For now all is working fine. As I mentioned in my most recent post in #33 I am able to avoid some of the memory conflicts by using your config to disable **RaiseInventoryAlloc.**

I'd love to chat sometime about the DEBUG. I'm always on Discord. 

Cheers

PR #34, opened 2022-01-14
`AshleyJPCameraAngles`: unlocks classic camera during Ashley segment

emoose:
This is done with a MakeInline hook (at 0x6F5518 in 1.1.0, sig also works for 1.0.6), so that it can be toggled while in-game fine (hopefully game won't have any issue with this getting switched about...)

I haven't really used MakeInline much before, so there's likely a better way to handle this.

Tested with 1.1.0, 1.0.6 & 1.0.6 debug, would be a good idea to check if JP version still works too though.

If anyone has a better name than `UnlockJPOnly` feel free to post it too!

### DIFF
--- a/dllmain/Settings.cpp
+++ b/dllmain/Settings.cpp
@@ -219,11 +219,12 @@ void Settings::ReadSettings()
 	cfg.bRememberWindowPos = iniReader.ReadBoolean("DISPLAY", "RememberWindowPos", false);
 	cfg.b60fpsFixes = iniReader.ReadBoolean("MISC", "60fpsFixes", true);
 	cfg.bFixQTE = iniReader.ReadBoolean("MISC", "FixQTE", true);
+	cfg.bAshleyJPCameraAngles = iniReader.ReadBoolean("MISC", "AshleyJPCameraAngles", false);
 	cfg.bSkipIntroLogos = iniReader.ReadBoolean("MISC", "SkipIntroLogos", false);
 	cfg.bEnableDebugMenu = iniReader.ReadBoolean("MISC", "EnableDebugMenu", false);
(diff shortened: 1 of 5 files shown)


emoose:
Ah might have zero-flag editing in MakeInline working, 1sec

nipkownix:
Just tested with the JP version and everything seems to be working smoothly.
`UnlockJPOnly` sounds good to me, but is there anything else exclusive from the JP version that we might be able to unlock? If it is just this one feature, maybe calling it something like `AshleyJPCameraAngles` could be more descriptive.

emoose:
I did see `pSys->languageId_8` checks near code that was loading in some textures, so could be possible to swap out logos like you mentioned - maybe some would only want to enable Ashley scene without changing anything else though, so might be a better idea to split them up...

I'll rename it in a sec, haven't really looked into the texture stuff yet though, maybe later on.

emoose:
Name updated, also fixed a small bug in case zero_flag was already set by something previously.

nipkownix:
Awesome. Thank you!

emoose:
btw about the Biohazard logo, looks like the call to `setLangExt3` at `0x6DE5BF` (1.1.0) handles that, seems that overwrites the beginning of `aTitleDat` with the language name, so it ends up as eg. `eng\title.dat`, editing that string to jpn straight after the `setLangExt3` call seems to get it showing Biohazard fine - maybe could just hook that call and do something similar.

nipkownix:
Ah, you're right.
It looks like it breaks some stuff though. The Separate Ways title screen had some weird letters on top of it, and the title menu itself had the japanese names for everything, such as Assignment Ada being called "Ada the Spy".
Maybe it isn't possible to change just the logo after all.

emoose:
Darn, I noticed the New Game difficulty selection was acting weird too, had two extra options that were both selected at once :S

Too bad, guess it'd need some way to edit title.dat (or the code that actually reads the logo from there)

Issue #35, opened 2022-01-14
4GB-patch indicator


emoose:
Might be a good idea to indicate to user whether they have 4GB patch active or not somehow, like the other DLL does, been a few times now that I've ran an EXE that wasn't patched & had issues because of it >.>

Doesn't seem like there's any way to 4GB patch it at runtime unfortunately - I wonder if we might be able to edit the EXE itself if the user hasn't already patched it though, hmm...

(also the way the other DLL checks for 4GB patch is wrong, seems they're checking for a constant 0x122 Characteristics value, when they should only be checking for the 0x20 bit, if you have the Relocations stripped flag active that adds 0x1 to it, so it's 0x123, making that DLL think it isn't 4GB patched 😆)

E: did a quick experiment, looks like on Windows you can't open the running EXE with fopen for writing, but you can move it!
Probably wouldn't be too hard to just copy the existing EXE + edit the Characteristics flags, then move them around so the patched EXE will be launched by user next time.

I'm not sure if being in Program Files could mess that up at all though, pretty sure I heard Steam sets up permissions on SteamApps dir to allow changes, but not 100% sure about that.

emoose:
Alright, got something working at https://github.com/emoose/re4_tweaks/tree/laa-check:

![20220115041548_1](https://user-images.githubusercontent.com/1755499/149608488-10dab159-242e-40d1-980d-4e244c305ce5.jpg)
![20220115041554_1](https://user-images.githubusercontent.com/1755499/149608489-f2a894cd-b6a8-47cf-b539-68cc800ca5bb.jpg)

Not sure if the way I implemented it is the best though, haven't really worked with ImGui that much before.
The code that handles the EXE patching should be pretty safe, tries restoring original EXE if we can't move the patched one in its place, so user isn't left with a broken install - probably more checks should be added there though.

nipkownix:
>(also the way the other DLL checks for 4GB patch is wrong, seems they're checking for a constant 0x122 Characteristics value, when they should only be checking for the 0x20 bit, if you have the Relocations stripped flag active that adds 0x1 to it, so it's 0x123, making that DLL think it isn't 4GB patched 😆)

Oh Raz0r.. 
If he had, you know, open sourced his DLL, we could have opened a PR with a fix or something 😆

> E: did a quick experiment, looks like on Windows you can't open the running EXE with fopen for writing, but you can move it!

Really? That's pretty interesting. I wonder if this behaviour is the same in previous Windows versions.

[...]

PR #37, opened 2022-01-16
Check if EXE is 4GB patched, prompt user & allow patching if not

emoose:
As mentioned at #35 

I changed the dialog titles to use 4GB patch instead of LAA patch since people are probably more familar with the 4GB name, added some extra line-breaks in, & also added the `ImGui::SameLine()` fix to the button positions.

22-01-16_21-15-43_bio4
22-01-16_21-15-49_bio4

Also did a quick test of removing my user permissions from the bio4.exe first before patching - interesting thing is that when I tried running game afterwards, the patch still worked fine! Looks like something added all the permissions back to the EXE after I ran it...

Guessing that's Steam related since game was in SteamApps folder, if I tried with a copy of the game in another folder then removing permissions gave a more expected result (excuse weird font in this pic, for some reason when I run 1.0.6 it scales the window weirdly for some reason)

22-01-16_21-07-20_bio4

(I was hoping to move buttons to right side too so they'd look closer to actual Win32 msgboxes, but couldn't find a way to anchor them cleanly, only way I saw was to specify X position directly, which wouldn't be good if dialog ever needs resizing...)

### DIFF
--- a/dllmain/ConsoleWnd.h
+++ b/dllmain/ConsoleWnd.h
@@ -1,6 +1,5 @@
 #pragma once
 
-
 struct ConsoleOutput
 {
 	void ShowConsoleOutput();
--- a/dllmain/LAApatch.cpp
+++ b/dllmain/LAApatch.cpp
@@ -1,0 +1,157 @@
+#include <iostream>
+#include "..\includes\stdafx.h"
+#include <d3d9.h>
+#include "dllmain.h"
+#include "Settings.h"
+#include "ConsoleWnd.h"
+#include "LAApatch.h"
+#include "..\external\imgui\imgui.h"
+#include "..\external\imgui\imgui_impl_win32.h"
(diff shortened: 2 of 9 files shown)


nipkownix:
Nice! Thank you.
I moved the new code to a new cpp file, made the buttons bigger and moved them to the right:

image

Looks more like Win32 msgboxes now. Let me know what do you think about these changes.

`(excuse weird font in this pic, for some reason when I run 1.0.6 it scales the window weirdly for some reason)`

I did notice that too. Might be worth taking a look at what causes that.

emoose:
Looks good 👍 Only one thing I can think of now, is there a way to lock the window so it can't be resized? Can't really think of a reason to keep it resizable anyway.

nipkownix:
Ah, forgot to mention it, but I did lock it.

PR #41, opened 2022-01-21
Wrappers: add `xinput1_3` & `x3daudio1_7` wrappers

emoose:
Both seem to be working with 1.1.0 at least, had to replace the `dsound` ordinals in Wrapper.def since xinput1_3 also requires fixed ordinals in some cases, game doesn't use dsound anyway afaik.

Using xinput1_3 might cause some issues with controllers for some people though, never had issues with it myself, but had some reports about that before (maybe they were using controller mapping software or something like that), x3daudio1_7 should be safer though since there's not much that mess with game audio AFAIK.

E: oh there is https://github.com/kosumosu/x3daudio1_7_hrtf that would conflict with x3daudio1_7 though... unfortunately that also works via the X3DAudio exports like `X3DAudioInitialize` etc, so chainloading with LoadLibrary probably wouldn't help there.

Not sure how much work it'd be to add in, but might be nice if maybe the wrapped DLL could be specified in INI instead of always using system DLL, so that we could load eg. x3daudio1_7_hrtf.dll and pass exports over to it - some mod DLLs use their DLL filename to find the system DLL though, so could cause issues with some things...

E2: hm, looks like we have a `RealDllPath` in dllmain.cpp that might allow for something like this, but nothing sets it up atm - seems INI is only read from after we've loaded the DLL though, so would probably need to reorder some things to allow customizing that...

### DIFF
--- a/Wrappers/wrapper.def
+++ b/Wrappers/wrapper.def
@@ -3,21 +3,18 @@ EXPORTS
 
 
 ; ****
-;LIBRARY "dsound"    <--- Needs to have the correct ordinals since some programs use the ordinal numbers for dsound APIs
+;LIBRARY "xinput1_3"    <--- Needs correct ordinals, some games don't include import names for this some reason
 ; ****
-DirectSoundCreate				@1
-DirectSoundEnumerateA			@2
-DirectSoundEnumerateW			@3
+DllMain				@1
+XInputGetState			@2
+XInputSetState			@3
+XInputGetCapabilities	@4
+XInputEnable			@5
+XInputGetDSoundAudioDeviceGuids		@6
+XInputGetBatteryInformation	@7
+XInputGetKeystroke	@8
 DllCanUnloadNow					PRIVATE
 DllGetClassObject				PRIVATE
-DirectSoundCaptureCreate		@6
-DirectSoundCaptureEnumerateA	@7
-DirectSoundCaptureEnumerateW	@8
-GetDeviceID						@9
-DirectSoundFullDuplexCreate		@10
-DirectSoundCreate8				@11
-DirectSoundCaptureCreate8		@12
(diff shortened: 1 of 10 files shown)


emoose:
Fixed up the RealDllPath stuff so it'll use a path from INI instead, also made it check that the DLL actually exists first.

That should allow a kind of 'chain-loading' of other DLLs that share the same name, hopefully the DLL being chain-loaded will then be able to wrap the actual system DLL.

Haven't tested this too much yet, VS debugging does show that x3daudio1_7_hrtf.dll does get loaded in, followed by the system DLL being loaded by it, but not sure if HRTF actually works or not.

Hopefully this might help with raz0r DLL as well, didn't try that yet though.

nipkownix:
Ah, I was about to do something like that before my HDD decided to die on me, lol
Thanks once again! The new wrappers seem to work nicely here.

Chainloading raz0r's dll appears to not be possible, though. I've tried a few different approaches, but it either nothing happens (wihout vs debugger), or it crashes when loading (with vs debugger attached).

Meh. I guess anyone who wants to use raz0r's thing can just rename re4_tweak's dll to winmm or xinput1_3. Should work just fine. I do think goind for dinput8 as a default is best for us (and albert) seeing as how unreliable winmm seems to be.

I guess I can merge this now, unless you have some other idea about how to handle raz0r's dll.

emoose:
Should be fine to merge, haven't had any luck with the raz0r DLL, my guess is that it might be using its own filename to try finding system DLL with, but since we have to rename it for ours to load in then that won't work for it, too bad.

Yeah dinput8 should be fine, maybe should mention in README.md about how it can be renamed to work with other dinput8 wrappers, and the different filenames it can be renamed to.

PR #49, opened 2022-01-30
FixDisplayMode: use alternate codepath for res list populating func (fixes #48)

### LINKED ISSUE #48: Crash opening options from main menu
Using latest 1.7.2 release with the included dinput8.ini & clean 1.1.0 bio4.exe, game crashes when opening options menu - removing dinput8.dll allows it to open it fine though, so don't think it's an issue with my games config.ini or anything.

Log file:
[bio4.exe.20220129221113.log](https://github.com/nipkownix/re4_tweaks/files/7964858/bio4.exe.20220129221113.log)

Seems to be something to do with the resolution chooser again, I used NV control panel to add 90Hz option for my resolution, so I have both 90Hz & 60Hz for the same res, maybe that's something to do with it (my desktop is at 90Hz, but maybe it's only exposing 60Hz to D3D or something weird)

emoose:
Fixes options screen crash when desktop refresh rate doesn't match any D3D modes, original code would try filtering res list to specific refresh rate, but this alternate codepath seems to check for dupe width/height entries instead, not sure why they didn't use it.

Tested with 1.1.0, 1.0.6 & 1.0.6 debug.

### DIFF
--- a/dllmain/dllmain.cpp
+++ b/dllmain/dllmain.cpp
@@ -222,34 +222,12 @@ void Init_Main()
 			}
 		}; injector::MakeInline<DisplayModeFix>(pattern.count(1).get(0).get<uint32_t>(0), pattern.count(1).get(0).get<uint32_t>(6));
 
-		pattern = hook::pattern("8B 45 ? 83 F8 ? 7C ? 7E ? EB ? 8B 3B");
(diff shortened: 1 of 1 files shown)


nipkownix:
Looks good!
I'll test it with my virtual 144Hz display too just in case.

PR #63, opened 2022-02-03
FixDisplayMode: filter out display modes under 59Hz

emoose:
Doesn't seem to help with green-screen issue, seems the second EnumDisplayAdapters func we patch is only used for populating options menu, but this should help prevent some bad modes from being displayed at least.


nipkownix:
Looks good to me. Thanks.

Issue #54, opened 2022-02-02
Issue with missing buttons in game. About aspect ratio


M1sh2:
The problem is very old (from the release), it seems to be somehow related to the language that is used in the system. I noticed that it manifests itself if you start using, for example, the Russian language, the Taiwanese language, perhaps some other. The buttons disappear completely, although I use the English version, they are not there when the context menu appears.
Screenshots are attached.
And also, is it possible to correct aspect ratio for 4:3, 5:4, and for other non-standard resolutions? The old permissions even at that time were greatly curtailed by cinematic black bars.
![20220202181359_1](https://user-images.githubusercontent.com/59886778/152146097-8f981959-b064-4297-86d5-bd831860f665.jpg) 
![20220202181532_1](https://user-images.githubusercontent.com/59886778/152146479-317f3e2f-972e-495f-8810-17470e14e0ad.jpg)
![20220202181801_1](https://user-images.githubusercontent.com/59886778/152147604-67b42d6f-845d-4ba5-9e25-188600945667.jpg)
![20220202181601_1](https://user-images.githubusercontent.com/59886778/152147682-9361f5ba-93c4-4dbe-a034-6fed5ce203bc.jpg)

emoose:
Hmm that is weird - I remember something about C functions like printf relying on locale, causing stuff like %f to show as eg. 3.1415 in one locale while 3,1415 in others - wonder if overriding C locale on startup might help at all.

RiasatSalminSami:
If I run the game with MSAA, I can see that the game is rendered behind the black bars during both gameplay and cutscenes while playing on 4:3 aspect ratio.

That means if we can somehow remove the black bars from top and bottom on 4:3, it should work. Although it might show non-animated stuff during cutscenes behind those black bar areas, but gameplay should be fine.

nipkownix:
>E: tried using a hook to call games setlocale func before running WinMain, setlocale(LC_ALL, "ru-RU") didn't seem to have any effect unfortunately.

Might be because you actually need to have that language installed on windows, no? (not sure about how locale works on Windows, but that is how it works on Linux at least).

>That means if we can somehow remove the black bars from top and bottom on 4:3, it should work. Although it might show non-animated stuff during cutscenes behind those black bar areas, but gameplay should be fine.

Yeah, removing the black bars causes weird effects like that.
[...]

PR #64, opened 2022-02-03
FallbackToEnglishKeyIcons: falls back to English key icons for unsupported languages

emoose:
Seems to work across 1.1.0/1.0.6/1.0.6debug, it might be nice for re4_tweaks to activate this automatically depending on known-bad layouts (or maybe just always activate it, if no other layout besides English shows correctly?)

E: oh I missed the image you posted of all the available keys - I guess that shows that languages besides English must be supported then. Too bad we don't have any list of the ones that work, unless maybe it's all the in-game translations?

E2: updated so that languages that have translation will still be used, and made it enabled by default

Here's a build if anyone is able to test it out:
[re4_tweaks1.7.4-englishKeyFallback.zip](https://github.com/nipkownix/re4_tweaks/files/7991176/re4_tweaks1.7.4-englishKeyFallback.zip)

### DIFF
--- a/dllmain/KeyboardMouseTweaks.cpp
+++ b/dllmain/KeyboardMouseTweaks.cpp
@@ -37,6 +37,11 @@ LastDevice GetLastUsedDevice()
 	}
 }
 
+HKL __stdcall GetKeyboardLayout_Hook(DWORD idThread)
+{
+	return (HKL)MAKELANGID(LANG_ENGLISH, SUBLANG_ENGLISH_US);
+}
+
 void Init_KeyboardMouseTweaks()
 {
 	auto pattern = hook::pattern("A1 ? ? ? ? 85 C0 74 ? 83 F8 ? 74 ? 81 F9");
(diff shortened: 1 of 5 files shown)


nipkownix:
> E2: updated so that languages that have translation will still be used, and made it enabled by default

Good thinking. Incredible work as always.
Maybe we should wait for @M1sh2's test results before merging?

M1sh2:
> > E2: updated so that languages that have translation will still be used, and made it enabled by default
> 
> Good thinking. Incredible work as always. Maybe we should wait for @M1sh2's test results before merging?

I can check the changes only after 8 hours

nipkownix:
No worries, there's no rush ^^

M1sh2:
Снимок экрана (3)
Yes, it worked.
I checked, there are buttons in QTE, in the settings, in the inventory.

nipkownix:
Awesome! Thank you for all the help.

nipkownix:
@emoose Found a potential issue with this:

https://steamcommunity.com/app/254700/discussions/0/3108018050534635927/?tscn=1644762654

Not sure why it would make every prompt a 0.

emoose:
Hmm, only time I had 0 prompt appear was when I messed with how they converted keycodes to glyphs here: https://github.com/nipkownix/re4_tweaks/issues/54#issuecomment-1028474743

Think that might happen when `ToAsciiEx` fails to convert, or maybe when the game doesn't have a glyph for it. I just noticed `ToAsciiEx` does take in a HKL parameter, maybe that would need to be updated too.

emoose:
Someone using English language had the 0 key issue too, really weird: https://www.re4hd.com/?p=9471#comment-197509

nipkownix:
Hmm? Very strange. I'll try to find a way to reproduce the issue tomorrow.

nipkownix:
Could the game be passing an incorrect ``idThread`` to ``GetKeyboardLayout_Hook``? Or maybe the warning I fixed in ba6194ae7835 caused some rare issue?

emoose:
Game always seems to use idThread 0, so shouldn't be a problem there - can't really see anything wrong with that warning fix neither :/

Wonder if it could be caused by some English locale/KB layout being missing from users install, don't think Windows requires english KB layout to be installed, maybe if it's not available then ToAsciiEx doesn't return things properly? (the user on re4hd mentioned using English lang, but also mentioned being in Portugal, so maybe didn't have eng KB layout?)

If it is ToAsciiEx related I guess we could maybe dump what ToAsciiEx should return for everything game calls to it, and hook it up to use that data if it returns anything invalid, hoping there might be some easier way than that though...

emoose:
Hmm weird, seems Russian is broken even with the build in the OP now, AFAIK only thing different is using 1.0.1 patch with it, maybe Albert changed something in the key icon graphics which our fix doesn't work with?

Doesn't really make much sense though, since keys display fine with English, and all our fix does is try making game think that it's using English, hm..

PR #65, opened 2022-02-03
Add contributions section + mention IDA database

emoose:
Added a small section to README that mentions the IDA database, might attract more people to help out if they knew we filled in so much already.


nipkownix:
Good idea. I probably should've thought about doing this earlier.

PR #68, opened 2022-02-03
Remove savegame SteamID checks, allows easier save transfers, fixes #67

### LINKED ISSUE #67: Allow savegame transfers
Game seems to check for SteamID inside savegame00.sav file, and refuses to load if it's different... Seems to have an ID at 0x1E40, and I guess there's a checksum/hash somewhere that it checks too since editing that ID doesn't seem to work.

Probably wouldn't be too hard to remove SteamID check.

E: seems to be two checks inside `CRCVerifyCRC`, 0x83DB95 & 0x83DBA4, with those two nopped it seems to load saves from other people fine.

emoose:
Tested with 1.1.0/1.0.6/1.0.6debug, should allow swapping out savegame00.sav for other peoples fine.

Seems the steamID inside the save is also used as part of the CRC calculation, but that should be unaffected, loading from other persons save & making new save seems to rewrite the SteamID part at 0x1E40 with the users one instead.

Fixes #67

### DIFF
--- a/dllmain/dllmain.cpp
+++ b/dllmain/dllmain.cpp
@@ -138,6 +138,11 @@ void Init_Main()
 
 	Init_ExceptionHandler();
 
+	// Remove savegame SteamID checks, allows easier save transfers
+	pattern = hook::pattern("8B 88 40 1E 00 00 3B 4D ? 0F 85 ? ? ? ?");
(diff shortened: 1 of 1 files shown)


nipkownix:
Awesome! Didn't think it would be this simple, really. Thanks!

PR #71, opened 2022-02-03
FixDisplayMode: use EnumDisplayDevices to get refresh rate of monitor 0

### LINKED ISSUE #51: Game screen turns green
My screen turns green after following directions. It was normal before doing it. I have an msi 65 stealth with a 2070 graphics card i9 i believe. Please help. Conment on my twitter so i can see the response. Thanku

https://twitter.com/UnMetalAlex/status/1488669301464834048?t=Td8X-hUn7ilUOWiQJ1L-Ug&s=19

emoose:
Broke: letting game filter display modes to only 60Hz ones
Woke: using `EnumDisplaySettings` with NULL device name to get desktop refresh rate
Bespoke: using `EnumDisplayDevices` to get device name of adapter 0 first, to use with `EnumDisplaySettings`

This seemed to fix the green screen issue, I guess EnumDisplaySettings with NULL device was fetching from wrong monitor.

Hopefully `EnumDisplayDevices` device indexes will match up with D3D's, so maybe in future we can use the adapter number from games config.ini with it (right now theres a bunch of places where game ignores adapter number though and hardcodes adapter 0, eg. `D3D_FillDisplayModeVector` - might need to hook the D3D calls they're making, or reimplement the whole caller,  shouldn't be too hard to track them down though since they all use g_IDirect3D9 / Direct3DCreate9 afaik)

Fixes #51

### DIFF
--- a/dllmain/dllmain.cpp
+++ b/dllmain/dllmain.cpp
@@ -192,23 +192,28 @@ void Init_Main()
 	// Hook func that runs EnumAdapterModes & checks for 60Hz modes, make it use the refresh rate from Windows instead
 	if (cfg.bFixDisplayMode)
 	{
-
 		if (cfg.iCustomRefreshRate > -1)
 		{
 			g_UserRefreshRate = cfg.iCustomRefreshRate;
 		}
 		else
 		{
-			DEVMODE lpDevMode;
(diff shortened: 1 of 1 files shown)


nipkownix:
Nice! Not sure why I was under the impression that calling with NULL would default to adapter 0.
Many thanks for figuring this out.

PR #86, opened 2022-02-05
Mention some of the newer features in README.md

emoose:
Before we forget about them :p


laymonage:
Might worth adding a note about `CrashDumps` as well?

emoose:
Oh yeah good thinking, will update with it now.

nipkownix:
Nice! Merged. Thanks!

Issue #83, opened 2022-02-04
[Feature request]  Modifier key to switch off new mouse controls


RiasatSalminSami:
The idea is very simple. Say if I hold the modifier key and move the mouse, then the camera will rotate around Leon like it usually does in the default game. But when I release the key, then moving the mouse will just rotate Leon.

This can be applicable in both ways:

-If new mouse controls are enabled: Then the modifier key will switch the mouse controls to default vanilla controls.
-If new mouse controls are disabled: Then the modifier key will switch the mouse controls to new one.

abi-21:
Sounds like a good idea, similar like in Arma games where you hold ALT key.

emoose:
Good idea, I think for this it'd probably be good if we could support mouse buttons as well, don't think we have anything for them right now though.

RiasatSalminSami:
> Good idea, I think for this it'd probably be good if we could support mouse buttons as well, don't think we have anything for them right now though.

Yeah, instead of modifier key, we should be able to use the side mouse buttons or the scroll wheel button as modifier if possible.

emoose:
Had a try at implementing this here: https://github.com/emoose/re4_tweaks/tree/mouseturn-modifier

Seems to work, but camera acts strange after releasing modifier, somehow it stays locked at the position you move it to with modifier held, and doesn't center or anything when mouse-turning, is that something to do with the mouse-turn impl?

Here's a build if anyone wants to try it:
[re4_tweaks1.7.4-MouseTurningModifier.zip](https://github.com/nipkownix/re4_tweaks/files/8005886/re4_tweaks1.7.4-MouseTurningModifier.zip)

You should be able to edit the MouseTurnModifier key to mouse buttons too, MOUSE1 to MOUSE5 (or LMOUSE/MMOUSE/RMOUSE)

RiasatSalminSami:
> Had a try at implementing this here: https://github.com/emoose/re4_tweaks/tree/mouseturn-modifier
> 
> Seems to work, but camera acts strange after releasing modifier, somehow it stays locked at the position you move it to with modifier held, and doesn't center or anything when mouse-turning, is that something to do with the mouse-turn impl? I guess mouse-turning probably assumes camera is centered at all times.
> 
[...]

PR #87, opened 2022-02-05
MouseTurning: add modifier key-combo to temporarily disable turning

emoose:
From discussion at #83

Using WndProc to get the states would be a lot better, atm there's an issue with bMouseTurnStateChanged where that's only valid for the first `MouseTurnEnabled()` call after state changed, after it's been called once it'll end up resetting the state changed bool at the next call to it (so if that's used more than 1 time in a frame it'd get wrong info), WndProc should let us set that up seperately to all the MouseTurn calls though.

E: oh I missed @RiasatSalminSami's suggestion in that thread:
>-If new mouse controls are enabled: Then the modifier key will switch the mouse controls to default vanilla controls.
>-If new mouse controls are disabled: Then the modifier key will switch the mouse controls to new one.

That's pretty smart actually, so modifier key just inverts the `UseMouseTurning` setting pretty much, and users could still disable it all by clearing key combination (is ALT used by anything in vanilla game btw? would be nice to have that as default)

### DIFF
--- a/dllmain/ConsoleWnd.cpp
+++ b/dllmain/ConsoleWnd.cpp
@@ -7,6 +7,7 @@
 #include "..\external\imgui\imgui.h"
 #include "..\external\imgui\imgui_impl_win32.h"
 #include "..\external\imgui\imgui_impl_dx9.h"
+#include "WndProcHook.h"
 
 ConsoleOutput con;
 
@@ -14,40 +15,22 @@ ImGuiTextBuffer     Buf;
 ImVector<int>       LineOffsets; // Index to lines offset. We maintain this with AddLog() calls.
 bool                AutoScroll = true;  // Keep scrolling if already at the bottom.
 
-std::vector<KeyBindingInfo> ConsoleKeyInfoVector;
-
 std::string TitleCombo;
 
-// Reads the key combo into an array containing the key id and isPressed state
+std::vector<uint32_t> ConsoleCombo;
+
 bool ParseConsoleKeyCombo(std::string_view in_combo)
 {
-    ConsoleKeyInfoVector.clear();
-
-	std::vector<uint32_t> ConsoleCombo = ParseKeyCombo(in_combo);
-
-	for (auto& key : ConsoleCombo)
-	{
-		KeyBindingInfo curkey = { key, false };
-        ConsoleKeyInfoVector.push_back(curkey);
-	}
-
-	return ConsoleKeyInfoVector.size() > 0;
+    ConsoleCombo.clear();
+    ConsoleCombo = ParseKeyCombo(in_combo);
+    return ConsoleCombo.size() > 0;
 }
 
-void ConsoleBinding(UINT uMsg, WPARAM wParam)
+void ConsoleBinding()
 {
-    if (IsComboKeyPressed(&ConsoleKeyInfoVector, uMsg, wParam))
+    if (IsComboKeyPressed(&ConsoleCombo))
     {
         bConsoleOpen = !bConsoleOpen;
-    }
-
-    switch (uMsg) {
-    case WM_KEYDOWN:
-        if (wParam == VK_NUMPAD3)
-        {
-            con.AddLogChar("Sono me... dare no me?");
-        }
-        break;
     }
 }
 
--- a/dllmain/ConsoleWnd.cpp
+++ b/dllmain/ConsoleWnd.cpp
@@ -16,6 +16,8 @@ bool                AutoScroll = true;  // Keep scrolling if already at the bott
 
 std::vector<KeyBindingInfo> ConsoleKeyInfoVector;
 
+std::string TitleCombo;
+
 // Reads the key combo into an array containing the key id and isPressed state
 bool ParseConsoleKeyCombo(std::string_view in_combo)
 {
@@ -203,7 +205,7 @@ void ConsoleOutput::ShowConsoleOutput()
     ImGui::SetNextWindowSize(ImVec2(500, 400), ImGuiCond_FirstUseEver);
     ImGui::SetNextWindowCollapsed(!bConsoleOpen);
 
-    std::string name = std::string("Console Output - ") + cfg.sConsoleKeyCombo + std::string(" to Show/Hide");
+    std::string name = std::string("Console Output - ") + con.TitleKeyCombo + std::string(" to Show/Hide");
     ImGui::Begin(name.data());
     ImGui::End();
     Draw(name.data());
--- a/dllmain/ConsoleWnd.h
+++ b/dllmain/ConsoleWnd.h
@@ -13,7 +13,7 @@ struct ConsoleOutput
 	void Clear();
 };
 
-void ConsoleBinding(UINT uMsg, WPARAM wParam);
+void ConsoleBinding();
 bool ParseConsoleKeyCombo(std::string_view in_combo);
 
 extern ConsoleOutput con;
--- a/dllmain/ConsoleWnd.h
+++ b/dllmain/ConsoleWnd.h
@@ -2,6 +2,7 @@
 
 struct ConsoleOutput
 {
+	std::string TitleKeyCombo;
 	void ShowConsoleOutput();
 	void AddLogChar(const char* fmt, ...);
 	void AddLogHex(int fmt, ...);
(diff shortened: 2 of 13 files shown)


emoose:
Updated with @RiasatSalminSami's suggestion, here's a build:
[re4_tweaks1.7.4-MouseTurningModifier-v3.zip](https://github.com/nipkownix/re4_tweaks/files/8007450/re4_tweaks1.7.4-MouseTurningModifier-v3.zip)
(this includes latest commits like savegame transfer fix)

One thing I noticed, if you make the stuck-camera bug happen, but then mouse-turn character just a tiny bit, then press modifier + release, that seems to unstick camera a bit, so seems mouse-turn might be able to overwrite camera rotation for us, would just need to force that to happen when state changes maybe.

RiasatSalminSami:
First of all, thanks a lot for considering my suggestion :D


I tested this and these are the results so far:

Camera centres itself now which is great.

The 2nd problem still remains though. If I turned Leon left or right [using mouse] and then pressed the modifier key to disable mouse turning, the camera will rotate all the way to left or right.

But now that the camera centres itself after releasing the modifier key, it feels very usable now.

nipkownix:
I've been moving the GetAsyncKeyState stuff to WndProc, but I'm seeing something odd.

The ALT key is actually sending WM_SYSKEYDOWN/UP messages instead of the usual WM_KEYDOWN/UP, and the wParam is VK_MENU(0x12), which isn't present in our keymap. ALT seems to be the only key doing this werid stuff.

Another problem I'm having is the game doesn't receive any mouse messages. I think they are being disabled by whatever input method they're using for the mouse in-game. So no mouse bindings currently, sadly.

@emoose any idea why this could be happening?

nipkownix:
@RiasatSalminSami I've made some changes and updated this PR to the lastest code changes.

Here's a build:
[re4_tweaks1.7.4-MouseTurningModifier-v4.zip](https://github.com/nipkownix/re4_tweaks/files/8010564/re4_tweaks1.7.4-MouseTurningModifier-v4.zip)

Let me know what still needs fixing. (it does seem pretty solid here though)

Still have to figure out mouse key bindings though. Meh.

emoose:
Hmm, looks like whenever a key event is sent, this is checking with each component for them to check keys against their combo vectors? 

The way I thought about handling WndProc was to update an array for each KEYDOWN/KEYUP (and I guess SYSKEYDOWN too), then have the components check against that array when they need to check combos, I think with that we could probably use the existing RawMouse code to update the VK_LBUTTON/VK_RBUTTON etc states as well for us.

eg.
```C
struct key_state
{
  bool pressed = false;
  bool changed = false;
}

WndProcHook.cpp:
std::vector<key_state> vk_states(0x100); // are there VKs with value more than 0x100?

key_state& GetVKState(int vk) // also in WndProcHook.h for other stuff to use
{
  return vk_states[vk];
}

[..]
WndProc()
{
  case WM_KEYUP:
  // update key state map
  {
    bool new_state = false;
    vk_states[wParam].changed = vk_states[wParam].pressed != new_state;
    vk_states[wParam].pressed = new_state;
  }
  break;
  case WM_KEYDOWN:
  // update key state map
  {
    bool new_state = true;
    vk_states[wParam].changed = vk_states[wParam].pressed != new_state;
    vk_states[wParam].pressed = new_state;
  }
  break;
}
```
(I did have something like this working but ran into same issue with ALT not being detected, ended up deleting it because of that, should have looked up what WM was actually being used for it >.>)

nipkownix:
Oh, looks like we had very different ideas for it. 
Not sure which method is more efficient, though. Hmm.

>maybe with that we could use the existing RawMouse code to update the VK_LBUTTON/VK_RBUTTON etc states as well for us

Problem with the RawMouse stuff is that `RegisterRawInputDevices` is being run every frame, basically.
I've never used it before, but surely you're only supposed to register it once. I'm having to do it like this because, if I don't, the game acts as if I've never registered for raw input at all. Almost as if the raw input device was registered and immediately unregistered. 
This is probably terribly inefficient, but I couldn't get mouse wheel delta to use on the cfgMenu otherwise.

The game also acts weird when raw mouse input is being received. It acts like the "in-game" mouse doesn't exist. Probably because the game uses dinput for the mouse events, and when I register for raw mouse input, the game keeps waiting for some dinput mouse event that never arrives.

This raw mouse stuff is pretty complicated to get working externally like this.

I might uh, just use GetAsyncKeyState if the binding has a mouse VK in it.

A shame though. If we could get raw mouse stuff working reliably, I'm sure we could replace the current in-game mouse delta stuff and get rid of mouse acceleration pretty easily.

nipkownix:
Added mouse support by checking GetAsyncKeyState, after all. Couldn't find another way.

Build here:
[re4_tweaks1.7.4-MouseTurningModifier-v5.zip](https://github.com/nipkownix/re4_tweaks/files/8011338/re4_tweaks1.7.4-MouseTurningModifier-v5.zip)

@emoose If you think you can improve any of this input stuff, feel free to do so.

(also, are the camera issues @RiasatSalminSami mentioned still happening?)

RiasatSalminSami:
@nipkownix  yeah the camera problem still occurs here.  Let me upload some footage to demonstrate exactly what happens on my end. Will take a while since my net is slow, but I will update the post as soon as it uploads.

emoose:
^ to explain what happens there, it's looking like camera is still being moved while mouse-turning is active, but the updated camera isn't actually shown until mouse-turning is toggled off with modifier (probably because mouse-turning forces camera centering, seems that only centers the camera being shown though, while I guess there's probably separate rotation values somewhere that are always updated despite it)

At least that's how the issue looks like to me anyway.

emoose:
https://www.reddit.com/r/pcgaming/comments/slov79/reshade_502_have_been_released/
>Fixed input not working in Resident Evil 4

Huh, doesn't ReShade use Imgui too?

Not really sure how it was fixed though: https://github.com/crosire/reshade/commit/040edc0e1d82b5928aac8aaa5901c03903a607a7

emoose:
Nice, camera issue seems to be fixed with that.

>Problem with the RawMouse stuff is that RegisterRawInputDevices is being run every frame, basically.
>I've never used it before, but surely you're only supposed to register it once. I'm having to do it like this because, if I don't, the game acts as if I've never registered for raw input at all. Almost as if the raw input device was registered and immediately unregistered.

Hm, just tried making it only execute once and it seems to work fine on my end (just added a static bool check around it), wonder if you might have something running which is unregistering the device or something?

I was wondering if maybe Steam overlay could have been doing something that messes it up, but seems to be working fine for me, do you have any other kind of overlays installed?

RiasatSalminSami:
> I was able to clean up the key combo stuff thanks to emoose's code a few posts ago. Thanks again :p
> 
> I've also fixed the camera issue that was shown in the video. Not sure how I didn't notice that before. [re4_tweaks1.7.4-MouseTurningModifier-v6.zip](https://github.com/nipkownix/re4_tweaks/files/8017119/re4_tweaks1.7.4-MouseTurningModifier-v6.zip)
> 
> @RiasatSalminSami Let me know if it is working as intended.
>

Beautiful! 

I can't thank you enough.  This is exactly what I had in mind when I suggested this modifier idea. 

Thank you again!

nipkownix:
> MOUSE1/2/3 doesn't work

> mouse 1 and 2 buttons acts as forward and back button respectively

Hmm, just to make sure: you're not confusing mouse 1 and 2 with mouse 4 and 5, right?

Mouse 1: Left click
Mouse 2: Right click
Mouse 3: Middle click
Mouse 4: Side button 1
Mouse 5: Side button 2

My mouse's side buttons are broken atm, so I couldn't test them, but they should be working since Mouse 1 and 2 are working fine for me. Hmm.

RiasatSalminSami:
Yeah, my bad. I typed mouse 1/2 t thinking it was for the side buttons.

4/5 works for the side buttons.

nipkownix:
Cool. I'll go ahead and merge this, then.

>Wonder if maybe we could grab mouse state from the games Key/Joy structs, AFAIK Joy is only controller related, but Key does seem to hold key/mouse state.. haven't really mapped out the struct for that much yet though (seems a lot different to Joy)

Hmm, this would probably be better than the current method.

I'll still be looking to make raw input work like it does on ReShade though.
Feel free to open another PR if you have something related to input in mind, and thanks again for all the help you provided with stuff in this PR. It is greatly appreciated!

nipkownix:
Managed to properly implement raw input, same way as ReShade does it. Will probably create a PR later for others to test it and make sure everything is stable, since it will require some big changes.

PR #89, opened 2022-02-05
LAApatch: repair PE checksum (fixes #35)

### LINKED ISSUE #35: 4GB-patch indicator

emoose:
Hopefully fixes issue with Chinese version of game mentioned in #35

### DIFF
--- a/dllmain/LAApatch.cpp
+++ b/dllmain/LAApatch.cpp
@@ -23,6 +23,20 @@ bool LAApatch::GameIsLargeAddressAware()
 	return (nt_headers->FileHeader.Characteristics & IMAGE_FILE_LARGE_ADDRESS_AWARE) == IMAGE_FILE_LARGE_ADDRESS_AWARE;
 }
 
+//https://programmerall.com/article/8899839695/
+uint32_t calc_checksum(uint32_t checksum, void* data, int length) {
+	if (length && data != nullptr) {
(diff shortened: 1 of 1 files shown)


nipkownix:
Sweet. I'll send a build to albert so he can provide it to the people who were having this issue. Thank you.

PR #91, opened 2022-02-05
Adjust Inventory/Files/Map/Merchant/Radio text for 32:9 Aspect Ratio

ksavery:


ksavery:
This does not fix issues like the "Check/Kick" button text covering the interaction icon, and the font being scaled to a level to exceeds the chatbox

nipkownix:
Oh, nice! Thank you for your contribution.

I'll update the readme before the next release to mention 32:9 and the minor issues.

PR #98, opened 2022-02-09
RaiseInventoryAlloc: increase game heap sizes instead of pointing to our own buffers, fixes #58

### LINKED ISSUE #58: RaiseInventoryAlloc: Invisible models for unknown reasons
Example from one user:

image

Seems to be the same issue from a while ago. I sadly still can't reproduce this issue for some reason. Tried starting Mercenaries with wesker on all stages, but the silencer was always visible.

Gotta admit I'm stumped here. Hopefully @emoose could have some idea.

If anyone experiences this issue and can help us testing possible workarounds, please let us know.

emoose:
Seems to solve the invisible items bug, many thanks to qingsheng8848 for posting code for their fix! (and also to everyone that helped debug it with us too!)

Cleaned up code a little, removed pointers/hooks that are no longer needed etc.

Should probably spend some time testing game with these changes to make sure heap increases don't break anything else, if anyone wants to help try them there's a build available here: [dinput8-heapHacks-noConsole-clean.zip](https://github.com/nipkownix/re4_tweaks/files/8035322/dinput8-heapHacks-noConsole-clean.zip)

### DIFF
--- a/dllmain/HandleLimits.cpp
+++ b/dllmain/HandleLimits.cpp
@@ -9,22 +9,22 @@ void Init_HandleLimits()
 	if (cfg.bRaiseVertexAlloc)
 	{
 		auto pattern = hook::pattern("68 80 1A 06 00 50 8B 41 ? FF D0 85 C0 0F 85 ? ? ? ? 46");
-		injector::WriteMemory<int>(pattern.count(2).get(0).get<uint32_t>(1), 0x68000C3500, true);  // 400000 -> 800000
-		injector::WriteMemory<int>(pattern.count(2).get(1).get<uint32_t>(1), 0x68000C3500, true);
+		injector::WriteMemory<int>(pattern.count(2).get(0).get<uint32_t>(1), 800000, true);  // 400000 -> 800000
+		injector::WriteMemory<int>(pattern.count(2).get(1).get<uint32_t>(1), 800000, true);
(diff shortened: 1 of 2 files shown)


nipkownix:
Cool! Awesome work. I'll send a build to albert so he can help testing and see if the chinese issue are fixed as well. Thanks for all your work!

emoose:
^ last commit includes fix for #99

nipkownix:
Albert just reported that the Chinese font issues are gone too 🎉 (turns out it affected Japanese too)
(why did it only affect these languages tho? and why is the ``subScreen`` memory (ss_pzzl stuff) even related to that? Guess it is one of those mysteries we will never know the answer to.

nipkownix:
Guess we're safe to merge this now?

laymonage:
> Guess we're safe to merge this now?

I've been playing with this patch from Chapter 3-1. I'm in Chapter 5-3 now and haven't noticed any issues. I also have the tactical vest and Mafia Leon fixes (made a custom build to include those and this PR) since 5-1. No issues so far.

The silencer was invisible in the inventory and the Merchant until I applied this patch.

nipkownix:
Nice. Thank you, emoose.

Also thank you, @qingsheng8848 for sharing your code with us.

PR #102, opened 2022-02-10
Add GitHub Actions workflow to build the project

laymonage:
Thanks for all the hard work! I'm not familiar with C++ and reverse engineering, so I tried to find another way to contribute.

This adds a GitHub Actions workflow to automatically build the project (with both `Debug` and `Release` configurations) and upload both DLL outputs as an artifact. Might be useful to let other people test new commits and PRs, while also ensuring reproducible builds of the project.

I don't know if this setup is sufficient, though. I might be missing some options to `msbuild`.

We can also automate the release with something like [action-gh-release](https://github.com/softprops/action-gh-release), but I suppose you probably want to do it manually so you can test things out ~~(and add the `.ini` as it seems to be written manually?)~~

The result is something like this (you can download the artifact): https://github.com/laymonage/re4_tweaks/actions/runs/1822496969

Let me know if you want me to make adjustments or add some other steps to the build process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
 
       - uses: actions/checkout@v2
         with:
-          submodules: "recursive"
+          submodules: recursive
 
       - name: Add msbuild to PATH
         uses: microsoft/setup-msbuild@v1.1
@@ -33,10 +33,16 @@ jobs:
         run: |
           msbuild re4_tweaks.sln /p:Configuration=Release /p:PostBuildEventUseInBuild=false
 
+      - name: Create default configuration file
+        shell: bash
+        run: |
+          head -n -1 settings/settings_string.h | tail -n +3 > dinput8.ini
+
       - name: Upload
         uses: actions/upload-artifact@v2
         with:
           path: |
             Debug/dinput8.dll
             Release/dinput8.dll
+            dinput8.ini
           name: re4_tweaks_${{ github.head_ref || github.ref_name }}_${{ steps.date.outputs.date }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,42 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Build Debug and Release
+    runs-on: windows-latest
+
+    steps:
+      - name: Get current date
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d_%H%M%S')"
+
+      - uses: actions/checkout@v2
+        with:
+          submodules: "recursive"
+
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1.1
+
+      - name: Build Debug
+        run: |
+          msbuild re4_tweaks.sln /p:Configuration=Debug /p:PostBuildEventUseInBuild=false
+
+      - name: Build Release
+        run: |
+          msbuild re4_tweaks.sln /p:Configuration=Release /p:PostBuildEventUseInBuild=false
+
+      - name: Upload
+        uses: actions/upload-artifact@v2
+        with:
+          path: |
+            Debug/dinput8.dll
+            Release/dinput8.dll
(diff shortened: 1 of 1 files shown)


laymonage:
Oh I just noticed you already added an AppVeyor integration in the latest commit...

nipkownix:
Thank you! I've been thinking about looking into GitHub Actions for a while.

>Oh I just noticed you already added an AppVeyor integration in the latest commit...

Ah, that was just a test. I was trying to see how much work it would take to build the project using mingw-w64 in there, but quickly gave up.

> We can also automate the release with something like [action-gh-release](https://github.com/softprops/action-gh-release), but I suppose you probably want to do it manually so you can test things out (and add the .ini as it seems to be written manually?)

Yeah, usually I think releases are better done manually.
About the .ini, the default one is in `/settings/settings_string.h`, but it is inside a const char. Not sure how difficult/possible it would be to get only the actual settings and move them to an ini file during build. I know some people use lua or python scripts for that, but I havent looked into it.

> Let me know if you want me to make adjustments or add some other steps to the build process.

I think everything looks fine. @emoose any suggestions?

laymonage:
> Yeah, usually I think releases are better done manually. About the .ini, the default one is in `/settings/settings_string.h`, but it is inside a const char. Not sure how difficult/possible it would be to get only the actual settings and move them to an ini file during build. I know some people use lua or python scripts for that, but I havent looked into it.

Thanks. I added a small `awk` script to extract the string and put it in `dinput8.ini`. [It seems to work.](https://github.com/laymonage/re4_tweaks/suites/5250408180/artifacts/161764181) As long as it starts at line 2, column 37 and the last line ends with `)"""";` :laughing:

I'm not a shell script expert so that might not be the best solution, but that's much faster than having to set up Python on the runner.

laymonage:
Also, it seems that the console is enabled even when building with `/p:Configuration=Release` 🤔

laymonage:
> Yeah, it just adds and empty new line at the very top and bottom of the generated .ini file, but I went ahead and pushed some changes to make `WriteSettings` skip that line. The bottom one will still be there, but it is no big deal.

Great, thanks. The bottom one is actually not a problem because text files are supposed to end with an empty newline (though it's mostly no longer needed these days).
 
> Hmm, that's because we `#define VERBOSE` in `dllmain.h` by default.

How do you usually build for `Release`? Do you manually remove that line? If so, I added a step just before the release build to do just that.

> Not seeing that prompt, for some reason.

Hmm, then I guess new workflows can only be added by maintainers of the repo.

One thing I noticed is that you need to login to download the artifacts, so this is only directly useful for those who have a GitHub account. But I guess we can just attach the file to a comment if we need to share it publicly.

nipkownix:
> How do you usually build for Release? Do you manually remove that line? If so, I added a step just before the release build to do just that. Seems to work just fine.

Yeah, I just comment that line out, so we should be fine then.

>One thing I noticed is that you need to login to download the artifacts, so this is only directly useful for those who have a GitHub account. But I guess we can just attach the file to a comment if we need to share it publicly.

Yeah, true.
Thank you! Let me know if you want this merged now. Might want to wait for emoose's input, dunno.

laymonage:
> Yeah, true. Thank you! Let me know if you want this merged now. Might want to wait for emoose's input, dunno.

No worries. I don't have anything to add atm, but let's wait for emoose.

emoose:
Looks good to me, nice use of head/tail to make the INI too, If I'd tried it probably would have ended up with a 30+ line script lol.

nipkownix:
Uh oh, merged this right before leaving for class, but it looks like the build is failing :/

laymonage:
> Uh oh, merged this right before leaving for class, but it looks like the build is failing :/

Huh, weird, looks like we got unlucky. It seems that GitHub is in the process of gradually upgrading the `windows-latest` image from `windows-2019` to `windows-2022` (https://github.com/actions/virtual-environments/issues/4856#issuecomment-1034087279). The build seems to work on [Windows 2022](https://github.com/laymonage/re4_tweaks/runs/5149771950) but not [Windows 2019](https://github.com/laymonage/re4_tweaks/runs/5149770653) due to incompatible VS versions. Let me make the PR to use `windows-2022`.

PR #108, opened 2022-02-11
Use `windows-2022` build image on GitHub Actions

laymonage:
This should fix the build. Context: https://github.com/nipkownix/re4_tweaks/pull/102#issuecomment-1035674261


nipkownix:
Nice. Thank you again!

PR #109, opened 2022-02-11
AllowSellingHandgunSilencer: updates merchant table to add silencer (fixes #100)

### LINKED ISSUE #100: Allow selling silencer to merchant
RE4HD added the hidden silencer back to the game, but as albert pointed out on re4hd.com:

>you can’t sell it, just discard it. The sell option of this item is not in the game’s code unfortunately

I did notice a bunch of tables related to item prices in the code though, so might be possible to edit those and add silencer to the list - `MerchantRoomInit` points to most of the lists, seems to be different lists depending on room, those might just be items you can purchase though, but maybe selling is somewhere around there too.

Think I saw some guides on residentevilmodding forum about editing merchant tables in the EXE too - @Mister-Curious have you heard of anyone doing something like this before?

g_item_price_tbl (C16E90 in 1.1.0) seems to be the main price table, probably used for selling items too - not sure what the structure of it is yet though, if anyone has experience with it please let us know!

emoose:
Seems to work fine for me with 1.1.0/1.0.6/1.0.6dbg, I set the price for selling it as 15000 since this is a rare easter-egg kinda item, open to changing it if anyone has a better idea of the price for something like that though (maybe it should be even higher?)

Many thanks to @Mister-Curious for helping to find more about the merchant tables, Re-Play Games on residentevilmodding forums for [documenting](https://residentevilmodding.boards.net/thread/13838/price-items-exe?page=1&scrollTo=130190) the table format, and also @laymonage for providing a save to help quickly test things 😄

### DIFF
--- a/dllmain/Settings.cpp
+++ b/dllmain/Settings.cpp
@@ -304,6 +304,7 @@ void Settings::ReadSettings()
 	// MISC
 	cfg.sWrappedDllPath = iniReader.ReadString("MISC", "WrappedDLLPath", "");
 	cfg.bAshleyJPCameraAngles = iniReader.ReadBoolean("MISC", "AshleyJPCameraAngles", false);
+	cfg.bAllowSellingHandgunSilencer = iniReader.ReadBoolean("MISC", "AllowSellingHandgunSilencer", true);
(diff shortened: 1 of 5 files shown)


nipkownix:
Looks pretty nice! Thank you.

Issue #22, opened 2021-11-24
Physics / Animations 


nipkownix:
I'm opening this issue so we can track these better.

Albert mentioned some issues related to animations and physics:

- [ ] Missing facial expressions in-game (Leon's expression when he's hurt or dies should change)
- [ ] Physics not working inside the inventory screen (causes clipping issues with Leon's jacket and the absence of Ada's black ribbon in the inventory)
- [ ] Leon's physics when he's wearing the tactical vest (pl09) are gone (static hair).

About the inventory physics issue, albert mentioned it was a working in the bio4 demo (for the GC i guess), but they disabled it/removed for the final version.

I've been looking at this, and it appears there's a function for each character that controls these physics.
For Leon, it is `cPlLeon::moveCloth`.
Noping it's call makes it look like it does in inventory screen:

![image](https://user-images.githubusercontent.com/33870163/143248382-09789c7a-645f-4d64-8d6d-a111c1276240.png)

It is called by the `cPlayer::move`, which itself is inside `gameMainLoop`.
`gameMainLoop` isn't called while inside the inventory screen, so we get static cloth.
Not sure this would simply be fixed by having ``cPl**::moveCloth`` inside the inventory screen's main loop, but who knows? 

About the tactical vest stuff: I do know the game applies different physics/cloth parameters based on what player model you're using. It is not something store inside the model itself (in this case pl09), but instead some type of table inside the .exe.
That _other dll_ fixed the issue, as far as I can tell.

emoose:
Hmm, had a look at the demo (for reference seems it's the japanese Taikenban demo, the US demos inventory is all 2D without any character render)
Couldn't actually rotate the char to check if cloth was moving or not, but since cloth did move when inventory opened I guess that meant its active.

Took a look at the .DOL for that demo, eventually found `PlClothMoveLeon` at 0x8013E4C4, opened the inventory & set a breakpoint on that in Dolphin, and seemed to be called from 0x80ADC79C.

Unfortunately GC builds use .REL files which separate the code into different files, kinda like Windows DLLs.
Tried searching for files that had matching bytes but no luck with that, some bytes were probably overwritten when the REL was loaded.
Eventually found some bytes that did match though - seems it's being called from Sscrn.rel, at 0x8050EDB8 if loaded  with IDA into 0x80500000 (not sure why lower bits of that addr are so different to what Dolphin got, but comparing code that dolphin shows seems to confirm that's the spot)

[...]

PR #115, opened 2022-02-14
Add patch to allow physics on tactical vest

emoose:
Includes the patch mentioned at https://github.com/nipkownix/re4_tweaks/issues/22#issuecomment-980542000, patching both `cPlLeon::initCloth` & `cPlLeon::moveCloth` (thanks to @RiasatSalminSami for reminding us about it!)

I still think it's weird they included code to check for this costume specifically though, doesn't seem that accidental to me - but since the patch mostly seems to work fine I guess we should be good, if anyone does find a problem with it we can probably just add a check for whatever edge-case has issues with it instead.

If anyone wants to test this, a build with this + RaiseInventoryAlloc fixes is at https://github.com/nipkownix/re4_tweaks/files/8055010/dinput8-tacticalVestPhysics.zip

As noted at https://github.com/nipkownix/re4_tweaks/issues/22#issuecomment-1037799043 Raz0r's trainer seems to overwrite the area that this patches, undoing it every frame, not sure if that can be worked around at all - but that makes me wonder if there could be other patches of ours that it overwrites too...

### DIFF
--- a/dllmain/dllmain.cpp
+++ b/dllmain/dllmain.cpp
@@ -479,6 +479,15 @@ void Init_Main()
 		uint32_t* g_item_price_tbl_num = pattern.count(1).get(0).get<uint32_t>(0);
 		*g_item_price_tbl_num += 1;
 	}
+
+	// Allow physics to apply to tactical vest outfit
+	// Some reason they specifically excluded that outfit inside the physics functions, maybe there's an issue with it somewhere
+	// (Raz0r trainer overwrites this patch every frame for some reason, too bad)
+	{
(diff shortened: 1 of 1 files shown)


nipkownix:
>Raz0r's trainer seems to overwrite the area that this patches, undoing it every frame, not sure if that can be worked around at all 

Most likely we can't work around that. This seems to be part of his "physics.cfg" patch. The default physics.cfg (from his dll) looks like this:

https://pastebin.com/raw/c18AhjHh

Both his trainer and his dll have options to load physics.cfg, and the trainer probably has Leon's vest hardcoded. 
physics.cfg is a pretty neat idea, tbh. I'm just not sure why he had to write everything every frame.

PR #119, opened 2022-02-17
FixAshleyBustPhysics: fixes framerate difference on physics applied to Ashley

emoose:
As explained at https://github.com/nipkownix/re4_tweaks/issues/22#issuecomment-1042438416 there's two functions related to Ashley bust physics, one for player-controlled Ashley & one for follower Ashley, on GC these two funcs are pretty much identical, but on PC it seems they updated the `cSubChar` version to make use of delta-time to fix physics speed at different framerates, but didn't copy that fix over to the `cPlAshley` function for some reason.

This hook pretty much just recreates that delta-time code inside `cPlAshley`, making physics apply at same rate despite framerate being used.

In effect this seems to slow down the physics at 60FPS - making it match the 30FPS physics speed - though I'm not sure if this helps with the main issue people had with 60FPS Ashley physics (maybe they're mostly referring to follower-Ashley, which this patch doesn't affect)

If anyone wants to test this a build is posted at https://github.com/nipkownix/re4_tweaks/issues/22#issuecomment-1042476550 - this hook only affects player-controlled Ashley, follower Ashley should remain unchanged.

(tested against 1.1.0/1.0.6/1.0.6dbg, hook seems to work fine at least)

### DIFF
--- a/dllmain/60fpsFixes.cpp
+++ b/dllmain/60fpsFixes.cpp
@@ -47,4 +47,32 @@ void Init_60fpsFixes()
 				_asm {fmul vanillaMulti}
 		}
 	}; injector::MakeInline<AmmoBoxSpeed>(pattern.count(2).get(1).get<uint32_t>(0), pattern.count(2).get(1).get<uint32_t>(6));
+
+	// Copy delta-time related code from cSubChar::moveBust to cPlAshley::moveBust
+	// Seems to make Ashley bust physics speed match between 30 & 60FPS
+	{
+		auto pattern = hook::pattern("57 E8 ? ? ? ? 8B 06 8B 50 0C 00 1D");
+
+		// Hook struct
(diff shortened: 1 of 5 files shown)


nipkownix:
This looks pretty nice.
So the ballistics physics are only affected in the player Ashley? Not sure if you compared npc Ashley in 30 and 60 fps already, but I'm guessing you did.

emoose:
Nah haven't really checked NPC Ashley much yet, seeing as the NPC code uses the same deltatime stuff I'd guess it should act the same as patched playable Ashley though, with 30/60 both running at same speed, will try confirming it soon.

I'm not totally sure what the issue people were reporting on re4hd was though, could be completely unrelated to the physics speed for all I know - but at least we can see physics speed does change between 30/60, which this PR fixes.

Issue #121, opened 2022-02-17
Game turns all button prompts to "0" when FallbackToEnglishKeyIcons is enabled


plrusek:
The option is also enabled by default.

Turned on:
![image](https://user-images.githubusercontent.com/18425248/154533482-86d75aa3-14b0-4484-8cd2-987f3eca2331.png)

Turned off:
![image](https://user-images.githubusercontent.com/18425248/154533204-b5e36377-be17-4344-b6c1-60c7a33c4769.png)

emoose:
Interesting, didn't think game had support for key icons with Polish layout, maybe the invisible icons issue only happens with Russian KB for some reason.

Do you know if you have English keyboard locale installed on your system at all? The 0 key issue might be happening when there's no English layout installed, not totally sure about that yet though.

plrusek:
Only Polish keyboard locale.

bursoft-git:
Hi! Thank you to your awesome work!
I hawe this issue too/
My system is Win11, Russian language, RE4HD 1.1 with re4_tweaks 1.7.7.6
When i use English keyboard layout, game shows D in fast press QTE, as expected

![bio4_2022_08_11_20_19_46_021](https://user-images.githubusercontent.com/7475720/184196056-d4673342-2ec9-4d75-a861-3ef1a2455196.png)

But when i switch to Russian layout, game shows 0

![bio4_2022_08_11_20_20_39_508](https://user-images.githubusercontent.com/7475720/184196073-c19cc1e9-4a29-4296-9241-7ebfa28ba167.png)

nipkownix:
Could you please open dinput8.dll, set "VerboseLog" to true, re-open the game and then upload your dinput8.log here?

bursoft-git:
> Could you please open dinput8.dll, set "VerboseLog" to true, re-open the game and then upload your dinput8.log here?

Yes, of course, here it is
[dinput8.log](https://github.com/nipkownix/re4_tweaks/files/9310765/dinput8.log)

nipkownix:
@bursoft-git Does this only happen with QTEs? Are other prompts fine? (enter door, save game, etc)

bursoft-git:
> @bursoft-git Does this only happen with QTEs? Are other prompts fine? (enter door, save game, etc)

Save game, open doors, and other actions with E button are OK
I do not check other QTEs, this one first i found
Later i check different QTEs with A/D buttons

nipkownix:
[...]

PR #123, opened 2022-02-17
KeyIconFallback: copy known-good keyicon data instead of relying on game

emoose:
This replaces the games keyicon data with known-good values from US key layout, if users language isn't one that we know game supports (also added LANG_POLISH to the supported list since it seemed #121 has it working without needing fallback enabled)

This should allow the fallback to work without needing any English keyboard locales installed, hopefully getting rid of the 0 key issue 😄

Tested on 1.1.0/1.0.6/1.0.6dbg (haven't tried JP versions though) - seems to work fine with both Russian & Polish for me, without needing English-US to be installed.


nipkownix:
Oh, very clever solution. Nice work!

PR #132, opened 2022-02-19
Game.h: move common Game pointers/structs to single header

emoose:
Moved some of the more common pointers & game-related structs to a Game.h header, and made them accessible through getter-only `GlobalPtr()` `PlayerPtr()` etc funcs.

Also added structs like `GLOBALS` etc from my IDB here, so we can access the fields normally instead of needing to use ugly pointer offsets, these structs are really barebones atm, but lets us get rid of most of those pointer accesses at least.

(just hope those structs all match between different game versions, seems like most things get stored at the same offset at least, so don't think they'd differ _too_ much)

If you think anything else should be moved there too let me know, wasn't really sure how much to put in there.

### DIFF
--- a/dllmain/60fpsFixes.cpp
+++ b/dllmain/60fpsFixes.cpp
@@ -1,11 +1,11 @@
 #include <iostream>
 #include "..\includes\stdafx.h"
+#include "Game.h"
 #include "dllmain.h"
 #include "Settings.h"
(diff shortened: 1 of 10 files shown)


nipkownix:
Very nice. Can't think of anything else to include there ATM, but I'll do so if something comes to mind later.
Thank you!

PR #136, opened 2022-02-25
UseSprintToggle: allows toggling sprint instead of needing to hold it, fixes #130

### LINKED ISSUE #130: [Request] Sprint without holding key down
As the title says, just being able to sprint without holding a key down would be really nice. And when you stop holding forward, your character will stop sprinting. Maybe make it a toggleable option as others may like it? Keyboard only option of course.

emoose:
Allows sprint to stay toggled until the FORWARD key is released, the toggle can be enabled/disabled during runtime at will.

It might be nice if this could be extended so that pressing sprint key again could untoggle it, but guess that might need some extra code to track when KEY_RUN was released & then pressed again (or maybe not, maybe `Key.buttonsTriggered_20` / `Key.buttonsReleased_28` could be made use of instead, not sure)

I think this might also affect controller too but haven't tried it myself, if it works fine there I guess maybe should move it to a different category instead.

Thanks to nipkownix for actually getting the hook to work too, seems I need to brush up on ASM some more...

Tested with 1.1.0/1.0.6/1.0.6debug, seems to work fine across them all.

(btw if anyone used the hex patches in #130, make sure to undo them before using this, since it relies on searching for the original bytes there)

### DIFF
--- a/dllmain/KeyboardMouseTweaks.cpp
+++ b/dllmain/KeyboardMouseTweaks.cpp
@@ -429,6 +429,35 @@ void Init_KeyboardMouseTweaks()
 		Logging::Log() << __FUNCTION__ << " -> FixSniperFocus applied";
 	}
 
+	// Hooks to allow toggling sprint instead of needing to hold it
+	{
+		// pl_R1_Run first KEY_RUN check
+		pattern = hook::pattern("83 E0 01 33 D2 0B C2 74 ? 8B C1 83 E0 40"); // 7639EB
+		struct SprintToggleHook1
+		{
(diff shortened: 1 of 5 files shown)


nipkownix:
Seems pretty good.
I'll go ahead and merge it, then test whether or not it works with controllers too. If it does, I'll move it to MISC. There are some other minor things I want to move, so I'll just do everything at once.

Thank you!

emoose:
Looks like it does work with controller fine, pretty neat. I think untoggling with sprint key would help that a lot though, will try looking into that soon.

PR #50, opened 2022-01-30
60fpsFixes: add patch to improve framelimiter above 60FPS

emoose:
Continued from older PR #25

(with the new framelimiter & `UseDynamicFrametime` setting in latest re4_tweaks most of this PR is no longer needed, the code that updates the settings menu with new framerates might still be useful in future though, so for now will probably just leave this PR as a draft.
There are still many issues with `UseDynamicFrametime` though, seems a lot of the game code is set up to only work properly when the frametime is exactly the same between frames, maybe some way of decoupling the graphics rendering from the deltaTime var would be good to look into...)

---

Hooks the `ConfigReadINI` call so we can update frametime used for framelimiter based on the `variableframetime` variable - with this **it seems framerates above 60 can work without being too broken**, likely many issues to be found (and hopefully fixed) though of course.

If anyone wants to try it out, here's a build with this PR:
(E: latest build can be found at https://github.com/nipkownix/re4_tweaks/pull/50#issuecomment-1046060996)
[re4_tweaks1.7.3_framelimit-fix-v2b.zip](https://github.com/nipkownix/re4_tweaks/files/7965542/re4_tweaks1.7.3_framelimit-fix-v2b.zip)

(old)
[re4_tweaks1.7.3_framelimit-fix-v2.zip](https://github.com/nipkownix/re4_tweaks/files/7965505/re4_tweaks1.7.3_framelimit-fix-v2.zip)
[re4_tweaks1.7.3_framelimit-fix.zip](https://github.com/nipkownix/re4_tweaks/files/7965196/re4_tweaks1.7.3_framelimit-fix.zip)

In v2 the ingame settings now includes extra framerates (30/60/90/120/144) - if you want even higher you can edit your RE4 config.ini file.
Framerate should probably be a multiple of 30, but maybe others will work fine too, who knows.

You may need to set `DisableVsync` in re4_tweaks dinput8.ini if the new framerates don't work for you, or unset fullscreen mode in game settings (or in config.ini) - hopefully if your monitor supports the refresh rate natively it should just work though.

If re4_tweaks shows a message about framerate setting, set `IgnoreFPSWarning` to true in dinput8.ini (should already be set in the above ZIP)

If you notice anything weird please let us know about it here - most problems are likely things that were set for 30FPS timings, game probably also had the same issue at 60FPS too, but running at higher framerates makes things much more apparent (try using `variableframerate 200` if you want to go on a bug hunt :)

~~TODO~~ (both should be working in latest commit)
[...]

PR #137, opened 2022-02-26
FixTurningSpeed: makes backwards turn speed consistent between framerates

emoose:
This is mostly useful with high framerates (where backward turning happens almost instantly), but it does slightly improve 60FPS too, making it match up with the 30FPS speed - also seems like the camera angle when turning gets improved by it a little bit as well.

(kinda silly that they didn't include this themselves, they did update pl_R1_Walk/pl_R1_Run to make use of deltaTime, but didn't bother with anything else...)

This also patches `pl_R1_KlauserAttack` since that also made use of the same `cPlayer::SPEED_WALK_TURN` var, but not really sure how to test that out yet, if anyone can try it out would appreciate it!

Tested with 1.1.0/1.0.6/1.0.6dbg, you can see the difference pretty easily by toggling it in cfgMenu while turning backwards.

### DIFF
--- a/dllmain/60fpsFixes.cpp
+++ b/dllmain/60fpsFixes.cpp
@@ -50,6 +50,77 @@ void Init_60fpsFixes()
 		}
 	}; injector::MakeInline<AmmoBoxSpeed>(pattern.count(2).get(1).get<uint32_t>(0), pattern.count(2).get(1).get<uint32_t>(6));
 
+	// Fix character backwards turning speed
+	{
+		struct TurnSpeedSubtract
+		{
+			void operator()(injector::reg_pack& regs)
+			{
+				float cPlayer__SPEED_WALK_TURN = 0.04188790545f; // game calcs this on startup, always seems to be same value
+				float newTurnSpeed = GlobalPtr()->deltaTime_70 * cPlayer__SPEED_WALK_TURN;
+
+				if (cfg.bFixTurningSpeed)
+					_asm {fsub newTurnSpeed}
+				else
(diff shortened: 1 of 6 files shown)


nipkownix:
Oh, pretty neat! I had no idea the turning speed was different. 

>This also patches pl_R1_KlauserAttack since that also made use of the same cPlayer::SPEED_WALK_TURN var, but not really sure how to test that out yet, if anyone can try it out would appreciate it!

Is ``pl_R1_KlauserAttack`` Mercenaries' Krauser's special attack?

emoose:
>I had no idea the turning speed was different.
Yeah only noticed it because 200FPS with #50 made turning speed happen almost instantly, which meant there couldn't be any kind of deltaTime applied to it, so tried changing between 30/60FPS with CE & that showed there definitely was a difference there.

Like I mentioned in that PR going through game at 100+FPS might be a good way of finding other stuff that aren't being timed properly... from looking at different parts of code affecting `cModel::rotation` it seems a huge majority of things were left un-deltaTimed, not sure how much of that actually matters though (was mostly hoping to find what handles enemy turning speed, haven't had any luck with that yet though...)

>Is pl_R1_KlauserAttack Mercenaries' Krauser's special attack?
Ah could be, haven't really played through Mercenaries yet so not really sure 😅 , I'll try testing with that later.

nipkownix:
I've just added some stuff that was missing from ``LogSettings()`` without realizing you added some of it here too. Oops :p

I'll grab a CI build from here once it is done and test the Krauser stuff. I'm very curious to see what it affects in practice.

Issue #142, opened 2022-03-02
CrashDumps improvements


emoose:
Some ideas for improving crashdumps, something to look into if any of us get bored:

- ~~Could we include savegame into the crashdump? or maybe savegame is included in minidump already? (not sure if it'd be fully loaded into memory though)~~

- ~~Most people don't know how to create ZIP files or upload things to the net - maybe using a single-header lib like https://github.com/richgel999/miniz to automate creating a ZIP could be a nice option, would let us cut down on the size of dumps too~~

- ~~Maybe worth creating an issue template that mentions something like "If you're experiencing game crashes try creating a CrashDumps folder next to the game EXE and see if the crash happens again, if it does then send us the new contents of the folder you created", if we do use ZIPs we could mention that the ZIP can be drag+dropped to github directly too. Also could probably add some questions like game/HD project version, have you tried checking with SFV...~~

- Would be sweet if we could get symbol names working somehow, think I saw something about converting IDB to a PDB before, maybe something like that would be enough.

E: hmm, looks like `bio4.exe+E22330` is a pointer to malloc'd mem that holds the save game (after main menu "loading" screen has shown), buffer pointer is filled fine inside minidump too, but doesn't seem the buffer itself got included into it in the dumps I looked at (same with the crash dumps that windows itself made inside %LOCALAPPDATA%\CrashDumps too)
Not sure if there's a way we could get dynamically alloced mem included in the minidump, could bloat it up a lot if we do... maybe easier to just dump out contents of that pointer to a file.

PR #146, opened 2022-03-03
Crash dump improvements: gamesave writing, try zipping dump to single file

emoose:
Seems to work pretty well, the ZIP creation does add a small delay when crash happens, but seems pretty negligible (I set this to just use compression level 3, tried out 10 but that made delay far too long... 3 is pretty fine though, reduces dump size a lot already) - though I'm not sure what this might be like on weaker machines.

Like mentioned in #142 we should probably let users know somehow that the zipped crashdump can be simply drag+dropped to github (instead of them needing to ZIP it themselves, or host it on gdrive or anything like that...)

Tested across 1.1.0/1.0.6/1.0.6dbg, seems to work fine with them all.

There was one strange thing I noticed though, when I made it crash with this it'd only generate a ~20MB dump file, but using the old `dinput8-heapHacks-noConsole-clean` build would make a 170MB one instead.
It's not the changes here AFAIK since I also tried with them removed, still only a ~20MB dump, switching back to heapHack build then still made a 170MB dump too, so doesn't seem like a fluke - was something changed in minidump writing code? didn't notice anything at least...

E: made a start on bug report & feature request issue templates too, let me know what you think of them: https://github.com/emoose/re4_tweaks/issues/new/choose


nipkownix:
Nice work!

> was something changed in minidump writing code? didn't notice anything at least...

Yeah, nothing changed. Not sure what caused this. We'll probably have to bisect to find which commit caused this (if any?)

> E: made a start on bug report & feature request issue templates too, let me know what you think of them: https://github.com/emoose/re4_tweaks/issues/new/choose

Looks awesome! I've never even considered templates. Those will be a very good addition.

Issue #153, opened 2022-03-18
[Feature request] German low-violence version support (& decensor?)


emoose:
### Describe the feature you'd like to see added

Another suggestion from re4hd.com comments: https://www.re4hd.com/?page_id=104#comment-198725

Somehow I totally missed that there was a seperate German version of the game, SteamDB also lists it under the game depots, seems to have its own EXE & data files, no idea what's actually different though since it seems my Steam sub doesn't have access to that depot ;_;

Hopefully the data files match 1.0.6 & it's just some small EXE difference, my guess is the EXE is probably based on 1.0.6 since it was updated same time as main 1.0.6 depot, no telling how close it matches with the normal 1.0.6 EXE though.

In worst case, if the EXE is a lot different to normal 1.0.6 I wonder if we could include something like a delta-patch to overwrite it with the 1.0.6 version in-memory, I guess things like relocations would probably make that a little harder though... but maybe we could patch the on-disk EXE like the 4GB patch trick - assuming it is just the EXE that's different and the data files are the same anyway.

**If any owners of the German version could upload the EXE** & mail me a link it'd be appreciated, my email is on github profile page: https://github.com/emoose

emoose:
Also I've heard JP version also has similar low-violence changes, a JP user also let me know that they seem to have crashes when using re4_tweaks (they aren't using re4hd though, not sure if that makes a difference)

>> > I tried using re4_tweaks, but it crashes every 5 minutes or so when
>> > checking objects.
>> > (X button on controller or E key on keyboard)

Although they later told me about https://otsushiki.hatenablog.com/entry/2016/11/28/200122 which details how to remove LV on JP savegames, and apparently with that mod the crashes went away, pretty interesting, haven't seen that kind of save mod before.

Unfortunately I forgot to ask about crashdump before they managed to fix their issue with that savegame mod, will try seeing if they can recreate the crash for us though.

nipkownix:
Awesome research!

I'll close this since it is implemented in #155. Thanks again!

PR #155, opened 2022-03-23
ViolenceLevelOverride: allows overriding level of violence in the game

emoose:
Some builds like JP/GER have low-violence-only EXEs, this should allow those to have full violence enabled instead, similarly full-violence EXEs should now be able to act like low-violence ones too.

More details in #153

Tested across 1.0.6/1.0.6jpn/1.0.6dbg/1.1.0, seems to be updating the pSys->field_9 field fine, haven't been able to test with German release yet though since I don't have the EXE for it atm.

I'm not totally sure if all the violence censorship is removed in JP/GER however since I'm not that familiar with those versions.

**If someone with those versions wants to try testing**, here's a build of this PR with violence set to full: [re4_tweaks-violenceLevelOverride.zip](https://github.com/nipkownix/re4_tweaks/files/8330405/re4_tweaks-violenceLevelOverride.zip)
 
E: ah, I've been told German copy is now uncensored, so guess we don't need to worry about testing with that - it'd still be interesting to get a look at the old censored release though, if any German owners know how to use DepotDownloader, the depot for that is at https://steamdb.info/depot/254702/ - looks like it should still be accessible to German owners.

### DIFF
--- a/dllmain/Game.h
+++ b/dllmain/Game.h
@@ -349,7 +349,7 @@ struct SYSTEM_SAVE
   uint32_t flags_CONFIG_0;
   uint32_t flags_EXTRA_4;
   uint8_t languageId_8;
-  uint8_t field_9;
+  uint8_t violenceLevel_9;
   uint8_t field_A;
   uint8_t field_B;
   uint8_t soundMode_C;
--- a/dllmain/Misc.cpp
+++ b/dllmain/Misc.cpp
@@ -9,6 +9,33 @@
 
 static uint32_t* ptrGameFrameRate;
 
+void(__stdcall* setLanguage_Orig)();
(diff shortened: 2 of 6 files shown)


nipkownix:
Nice work, emoose.

> E: ah, I've been told German copy is now uncensored, so guess we don't need to worry about testing with that

Yeah, it happened when ver 1.1.0 was released, AFAIK. New German owners were given that new version instead of the old, censored 1.0.6.

I'll merge this now, since the likelihood of someone reading this having the old censored version is pretty slim.

Thank you!

emoose:
Ah just realized I forgot to add any kind of logs for this (like the ini settings logging), might want to add that in before any new release, in case this does cause any issue...

Issue #158, opened 2022-04-01
Keyboard controls for doing backflips with the escape boat


RiasatSalminSami:
### Describe the feature you'd like to see added

So far I've been trying various keyboard button combinations to see if I can flip the boat, but nothing worked. As far as I know, it's possible to do it with gamepad on GC and Wii port. Can't test gamepad on PC right now since I don't have one.

I've been trying to find some answers regarding this, but there isn't any documentation for this. This was the only lead: https://steamcommunity.com/app/254700/discussions/0/627456487064225218/  and even there, no one managed to find anything.

I think, just like the lack of controls to flip the items in your inventory, you cant flip the boat with keyboard by default.

emoose:
Only been able to find button-press checks inside `pl0ePathMove` so far, looks like that's only looking for `KEY_LEFT/KEY_RIGHT/KEY_FORWARD/KEY_BACK`, haven't found any other code related to jet-ski that checks buttons.

If anyone with gamepad could test if the combo in that steam link works or not it'd be appreciated, seems like it might have been stripped out entirely.

linkthehylian:
I remember trying to do this several times with keyboard and it never ended up working. I thought I was just doing it wrong but it looks like it might have actually never been properly implemented. All the times I've tried with a controller have worked just fine though, so I think it's safe to assume that it only recognizes that kind of input to be activated.

emoose:
Ahh right, thought I was probably mistaken about it being taken out.

Just had another look and seems `pl0e_R1_Jump` does have a `(Joy[0].buttonStates0_14 & JOY_RT) != 0 && sub_401528(1, 0x80)` check inside it, not sure how I missed that before 😅 

I'm not certain but maybe the `Joy[0].buttonStates` field is only set from gamepad, and doesn't get updated from keyboard at all, not sure though. 

Or maybe it's the sub_401528 call there which is stopping it from working, I'm not really sure what that sub checks for (GC/PS2/VR don't have that call at all), will try looking into that some more.

linkthehylian:
I had to play a bit to get to where you wanted. It should be save number seventeen. @emoose 
[savegame00.zip](https://github.com/nipkownix/re4_tweaks/files/8414731/savegame00.zip)

emoose:
> I had to play a bit to get to where you wanted. It should be save number seventeen. @emoose [savegame00.zip](https://github.com/nipkownix/re4_tweaks/files/8414731/savegame00.zip)

[...]

PR #162, opened 2022-04-07
JetSkiTricks: patch gamepad-only code to allow mouse/keyboard combos

emoose:
Hey hope all is well, here's a PR for the jet-ski patches mentioned in https://github.com/nipkownix/re4_tweaks/issues/158#issuecomment-1088311559

This just hooks `pl0e_R1_Jump` so we can check for a keyboard/mouse combo before the controller checks, the way I patch around the checks here is a little hacky, as they check for both RT & LT with two seperate test/jz instructions, seemed easier to make the hook return past those tests & skip them if combo pressed instead of needing to hook them both individually, maybe there's a better way to handle that though.

Tested with 1.1.0/1.0.6/1.0.6dbg so far, not sure about other releases, as long as those RT/LT tests are always 0x28 bytes it should be fine, hopefully none of the other versions changed that at all.

Here's a build of these commits, has some very minor improvements from the earlier build:
[dinput8-jetSkiTricks-v2.zip](https://github.com/nipkownix/re4_tweaks/files/8441378/dinput8-jetSkiTricks-v2.zip)

### DIFF
--- a/dllmain/KeyboardMouseTweaks.cpp
+++ b/dllmain/KeyboardMouseTweaks.cpp
@@ -22,6 +22,15 @@ static uint32_t* ptrMouseAimMode;
 
 int iMinFocusTime;
 
+std::vector<uint32_t> jetSkiTrickCombo;
+
+bool ParseJetSkiTrickCombo(std::string_view in_combo)
+{
+	jetSkiTrickCombo.clear();
+	jetSkiTrickCombo = ParseKeyCombo(in_combo);
+	return jetSkiTrickCombo.size() > 0;
+}
+
 LastDevice GetLastUsedDevice()
 {
 	// 0 = Keyboard?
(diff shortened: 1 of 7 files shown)


nipkownix:
Looks good! Many thanks!

I imagine it is working fine in 1.0.6 jp as well? If so, we can probably safely merge this before the new release.

emoose:
Yep JP seems to work fine too, also double checked the versions I mentioned in OP to make sure and they seem fine as well.

I'm still not sure which part of the `(Joy[0].buttonStates0_14 & JOY_RT) != 0 && sub_401528(1, 0x80)` was failing the check originally, pretty sure I've seen other parts of the game checking Joy.buttonStates for things without breaking KB support, unless RT/LT are just odd buttons which KB/M never set for some reason...

I'm guessing the `sub_401528` call is probably more responsible tho, only seems used in ~4 different places, one of them strangely being in Filter04's draw function... will have to try looking at it more at some point.

RiasatSalminSami:
This is amazing. Thanks a lot.

PR #167, opened 2022-04-16
AllowAshleySuplex: allows Ashley to suplex in very specific situations

### LINKED ISSUE #163: [Feature request] Let Ashley suplex enemies
### Describe the feature you'd like to see added

In the Gamecube specific release of Resident Evil 4. If you manage to stun a cultist with one of the doors resulting them in kneeling, it shows the suplex interact prompt that Leon is normally supposed to have.

https://user-images.githubusercontent.com/20933012/162860397-bcf3a591-31ff-4ab9-bdb3-9eb82e6ff733.mp4

I realize this is a bit of a silly thing to request. But it was fixed after the Gamecube release and it's no longer possible because of that and it would be cool to see it return.

Sorry if the audio is absurdly loud, I had to get footage from YouTube.

emoose:
This fixes #163, allowing player to bring back the Ashley suplex glitch, thanks to @linkthehylian for suggesting it & helping test changes too!

Video: https://twitter.com/linkthehyIian/status/1514764457578180643

Method: https://github.com/nipkownix/re4_tweaks/issues/163#issuecomment-1099741577

I was hoping that the patch could also act as a fix to allow Ashley to Kick instead (the way they added the Ashley-check to the suplex func meant Ashley wouldn't have any action at all for downed enemies...), but right now it seems this fix doesn't actually work for that... but at least this lets us control it at runtime though.

Maybe looking into `em10ActEvtSetKick` some more could let us get Kick working, seems to have a bunch of player-type checks inside it at least, I guess something there must be failing. (if anyone is able to post a save just before where Suplex is performed that would help a lot, should be able to use Debug Menu to open savegame screen)

Tested with 1.1.0/1.0.6/1.0.6dbg/1.0.6jpn, the hook::patterns are found for them all at least, haven't tried the Suplex out myself though, but @linkthehylian reported that it worked for enabling the glitch at runtime though, not sure what game version they used.

### DIFF
--- a/dllmain/LoggingInit.cpp
+++ b/dllmain/LoggingInit.cpp
@@ -177,6 +177,9 @@ void LogSettings()
 	sprintf(settingBuf, "| %-30s | %15s |", "SilenceArmoredAshley", cfg.bSilenceArmoredAshley ? "true" : "false");
 	Logging::Log() << settingBuf;
 
+	sprintf(settingBuf, "| %-30s | %15s |", "AllowAshleySuplex", cfg.bAllowAshleySuplex ? "true" : "false");
+	Logging::Log() << settingBuf;
+
 	sprintf(settingBuf, "| %-30s | %15s |", "DisableQTE", cfg.bDisableQTE ? "true" : "false");
 	Logging::Log() << settingBuf;
 	
--- a/dllmain/Misc.cpp
+++ b/dllmain/Misc.cpp
@@ -230,4 +230,37 @@ void Init_Misc()
 		injector::MakeNOP(pattern.count(2).get(0).get<uint8_t>(7), 2);
 		injector::MakeNOP(pattern.count(2).get(1).get<uint8_t>(7), 2);
 	}
+
+	// Patch Ashley suplex glitch back into the game
+	// They originally fixed this by adding a check for Ashley player-type inside em10ActEvtSetFS
+	// However, the caller of that function (em10_R1_Dm_Small) already has player-type checking inside it, which will demote the suplex to a Kick for certain characters
+	// Because of the way they patched it, enemies that are put into Suplex-state won't have any actions at all while playing as Ashley, while other chars are able to Kick
+	// So we'll fix this by first removing the patch they added inside em10ActEvtSetFS, then add checks inside em10_R1_Dm_Small instead so that Ashley can Kick
+	// (or if user wants we'll leave it with patch removed, so that they can play with the Ashley suplex :)
+	// TODO: some reason the below doesn't actually let Ashley use SetKick, might be player-type checks inside it, so right now this just disables Suplex if bAllowAshleySuplex isn't set
+	{
+		// Remove player-type check from em10ActEvtSetFS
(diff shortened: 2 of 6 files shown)


linkthehylian:
I used 1.1.0 to test, I would've used the patched JP 1.0.6 executable but figured it would be better to use an updated one instead.

> (if anyone is able to post a save just before where Suplex is performed that would help a lot, should be able to use Debug Menu to open savegame screen)

Here you go. It's the Chapter 3-4 save.
[savegame00.zip](https://github.com/nipkownix/re4_tweaks/files/8499318/savegame00.zip)

emoose:
> Here you go. It's the Chapter 3-4 save.
> 
> **Edit**: I forgot use the debug menu to save with Ashley in front of the door. Here's the new one. [savegame00.zip](https://github.com/nipkownix/re4_tweaks/files/8499385/savegame00.zip)

Thanks! So I've confirmed that SetKick func does get ran, but the prompt doesn't appear, so must be some check inside it which fails... 

I guess that means Ashley never has any Kick prompts during normal gameplay then? Patching it to allow Kick could have a chance of messing up other things I guess, but maybe we could add something so Kick is only shown in the situations that Suplex would, guess it could be an easter-egg or something :P

For all I know Ashley might not even have the anims to use it though, will try seeing if I can force it to show up...

linkthehylian:
> Seems like it might be buggy though, I'm pretty bad at getting the enemy to kneel, @linkthehylian could you try testing with it?
Use CE and change bio4.exe+6119A (byte) from 01 to 255, with that & AllowAshleySuplex disabled it should show the Kick action, but seemed kinda broken the 1 time I got it to work...

Oh it's broken alright lol. I'm wondering how she's able to even perform a suplex if that happens with the kick. It looks like she's playing her ladder climbing animation instead?

https://user-images.githubusercontent.com/20933012/163665733-183f44af-6e6e-4b8a-94a6-ea18bec8c796.mp4

emoose:
> Oh it's broken alright lol. I'm wondering how she's able to even perform a suplex if that happens with the kick. It looks like she's playing her ladder climbing animation instead? This game is full of surprises. 😄
> 
>  Resident.Evil.4.Biohazard.4.2022.04.16.-.03.09.29.02.Dvr.mp4

Darn, yeah pretty strange that they didn't give her a Kick anim but did add a Suplex one... Starter Leon model also does some weird thing if you force him to Suplex as well.

Could be that it's just using the wrong anim number for her or something though, would be interesting if someone that knows how to look at anims could check if she has one for kicking.

Anyway for now I guess this PR is fine, if Suplex isn't enabled it'll fall back to SetKick which will exit out if using Ashley, in future if Ashley kicking is figured out it shouldn't be too hard to add that in as well.

nipkownix:
Guess we're safe to merge this, then. Hopefully we can figure out the Ashley kicking in the future.

Thank you!

nipkownix:
Guess we're safe to merge this, then. Hopefully we can figure out the Ashley kicking in the future.

Thank you!

PR #185, opened 2022-05-13
DatTbl: add hooks to allow loose files to be side-loaded into DAT/EVD

emoose:
E: updated this PR to include a patch to enable English subtitles as well, the patch will only be applied if it finds loose replacement subtitle files in the game folder, since the default Eng subtitles are pretty much dummy placeholders.

E2: latest commit also enables the Subtitles option in the settings menu when using english lang, seems this defaults to Off, and toggling it seems to work fine with cutscene subtitles - hopefully there won't be any weird issues caused by patching this like the 90/120/144FPS setting-menu-patches had...

E3: found a way to fix the Saddler-causing-low-FPS bug, #188 goes into detail about that, just added a patch to force it to 4 loops instead of 0x14, and with that the FPS issue is gone - I didn't notice any visual changes from this, but if anyone does notice any difference please let me know!
(E4: above sadder fix is removed, found a more proper fix at #192, though root cause might be something wrong in HD project cloth data...)

---

This allows the game to read from loose files instead of using data inside the EVD/DAT, if the file being requested from EVD/DAT exists loosely, letting us easily override data inside these DAT containers without needing to repack everything 😄 

Raz0r's DLL has a similar ability (@Mister-Curious did a great writeup about that here: https://residentevilmodding.boards.net/thread/13696/introduction-raz0rs-dll?page=1&scrollTo=129349) - but seems that might need a config file setup in advance with each file path to side-load, but instead here we can just check if the files exist on disk or not.

This works by memory-mapping the file when game requests it from the DAT table (DAT tables seem to get used in a bunch of different file types, EVD/DAT/DRS...), the file also gets unmapped when the game decides to free the DAT table entry (or DAT table itself), so hopefully no kind of memory-leaks or anything from it.

As an example of what can be done with this, here's a modded r100s03.mdt subtitle file for the cutscene in the beginning house:
[r100s03-mod.zip](https://github.com/nipkownix/re4_tweaks/files/8685868/r100s03-mod.zip)

Just drop that into the BIO4\event\r100\s03\etc\ folder (folder might need to be created), and now the game should read from that file instead of using the data from EVD.
(make sure to use submod EXE files to allow English subtitles to show - those are linked at https://github.com/emoose/re4-research/issues/11 )

---

This still needs quite a bit of testing though, would appreciate anyone trying it out, even if you aren't modding any files/subtitles - always a chance the way it checks which files to side-load can increase load times in some way (didn't really notice much difference myself, but that could be because I'm using SSD...), please let me know your thoughts on it!

A build with this PR included can be found here: [dinput8-dat-sideloading_eng-subtitles_cutscene-release-fix_saddler-fix.zip](https://github.com/nipkownix/re4_tweaks/files/8695712/dinput8-dat-sideloading_eng-subtitles_cutscene-release-fix_saddler-fix.zip)

(old vers: [dinput8-dat-sideloading_eng-subtitles_noConsole-settingsFix.zip](https://github.com/nipkownix/re4_tweaks/files/8694239/dinput8-dat-sideloading_eng-subtitles_noConsole-settingsFix.zip) / [dinput8-dat-sideloading_eng-subtitles_noConsole.zip](https://github.com/nipkownix/re4_tweaks/files/8694203/dinput8-dat-sideloading_eng-subtitles_noConsole.zip) /  [dinput8-dat-sideloading.zip](https://github.com/nipkownix/re4_tweaks/files/8685861/dinput8-dat-sideloading.zip))

@Mister-Curious also mentioned in the Raz0r DLL thread that sometimes files inside the DAT might have strangely named filenames, even though the data matches up to a different normal filename instead - in those cases Raz0r also allows using a lookup.txt file to remap the filename first, maybe worth adding something similar here too.


emoose:
Found one small issue with this, in the movie browser the last Chapter 4 cutscene (Ada in boat) opens `r22as00.mdt` to read in the subtitles, but for some reason after the cutscene movie ends it doesn't free it (seems this is a vanilla bug too, so maybe loading/exiting that cutscene over & over could break unmodded game as well.)

I've traced it to a `(v6 & 0x100000)` flag check inside `EventMgr::Run` (0x6284B3 in 1.1.0), if that flag is set then it won't call into `EventMgr::DelEvt` to free the event data etc. Other cutscenes don't seem to have that flag set, but this Ada one does for some reason.
Patching the flag check out lets it free the event fine, but don't really think this is a proper fix, need to find why that flag is set in the first place...

emoose:
Seems to be an issue when side-loading certain model files with this, looks like game wants to update some pointers inside the model data, but being memory-mapped means writes won't work. Did a quick test of changing it to use malloc/free instead and seemed to work fine.

It's not really that important though since the main use of this atm is for MDT replacements, but I guess people would probably expect other types to be replaceable as well, will see about updating the PR soon.

emoose:
Model issue should be fixed now, also now allow loading in the MDTs for the opXX.udas files from loose files instead, MDTs should be placed in `BIO4\op\unpacked` folder.

Here's a build of this PR along with MDTs for english subtitles & fixed up radio call texts:
[bio4-110-englishSubs-fixedRadio-englishSubPatch.zip](https://github.com/nipkownix/re4_tweaks/files/8748460/bio4-110-englishSubs-fixedRadio-englishSubPatch.zip)

Many thanks to [linkthehylian](https://github.com/linkthehylian), [Zaphero1](https://github.com/Zaphero1), [nr1995](https://github.com/nr1995) & [laymonage](https://github.com/laymonage) for writing/contributing to these text fixes!

emoose:
Ah darn, looks like op03 did have the issue I guessed it would, luckily it's an easy fix, force pushed with the fix.

Here's a new build of this PR along with MDTs for english subtitles & fixed up english radio call texts:
[bio4-110-englishSubs-fixedRadio-englishSubPatch-v3.zip](https://github.com/nipkownix/re4_tweaks/files/8748596/bio4-110-englishSubs-fixedRadio-englishSubPatch-v3.zip)

nipkownix:
Looks amazing! Incredible work. 
Just a thought: Would it be a good idea to side-load the loose files from a new directory near our .dll instead? Something like ``Bin32/re4_tweaks/sideload/event/r***/s***/etc/r***s***.mdt``, for example. This would make uninstallation simpler, since we avoid messing with the BIO4 folder at all, and it just feels cleaner IMO.

emoose:
Oh that might be better, didn't think about the cleanup for it, I'll update it with that path shortly.

Zaphero1:
Thank you very much!
> Here's a new build of this PR along with MDTs for english subtitles & fixed up english radio call texts: [bio4-110-englishSubs-fixedRadio-englishSubPatch-v3.zip](https://github.com/nipkownix/re4_tweaks/files/8748596/bio4-110-englishSubs-fixedRadio-englishSubPatch-v3.zip)
This build seems to just crash for me with this error.
image

emoose:
Updated with the new path + added a fix for the JP exe, also made it close file handle straight after reading rather than keeping it active while the `FileData` struct for it was alive.

New build with the english subtitle MDTs can be found here:
[bio4-110-englishSubs-fixedRadio-englishSubPatch-v4.zip](https://github.com/nipkownix/re4_tweaks/files/8773159/bio4-110-englishSubs-fixedRadio-englishSubPatch-v4.zip)

Also forgot to mention, but I'm still a little unsure about the
```
// Fix R245EventS00/R22EEventS00 accidentally calling EvtReadExec with a do-not-free-after-use (0x40) flag
// (without this fix, those cutscenes will leak memory whenever they're played...)
``` 
patch which removes that flag, maybe movie viewer didn't need it, but the actual in-game level maybe depends on something from the evt still being in memory... probably fine to merge for now, but maybe need to go through that level sometime before the next release (probably gonna go through game again once the math reimpl stuff is added anyway)

BTW do you think we should add the MDTs into the repo somewhere? I'm not really sure where to put them for github actions to include it in the zip artifact yet...

nipkownix:
> BTW do you think we should add the MDTs into the repo somewhere? I'm not really sure where to put them for github actions to include it in the zip artifact yet...

Not sure about GitHub actions either, but we could include the MDTs in the external folder, maybe? No objections from me.

emoose:
Nice, added them under `externals\EnglishSubtitleData\Bin32\re4_tweaks\sideload\[...]` folder, took a little while but eventually got github actions to include them in the ZIP as well (under `Debug\re4_tweaks\` & `Release\re4_tweaks\`, it's the same data in both so doesn't really increase ZIP size much at all): https://github.com/nipkownix/re4_tweaks/actions/runs/2399956326

nipkownix:
Very nice! Thank you. I'll go ahead and merge it, while keeping a mental note to go through that level you mentioned before we do another release.

Issue #188, opened 2022-05-15
Lag during cutscenes caused by `PenClothMove3` / `penClothAtCkParallel`


emoose:
### Validation

- [X] Game has been updated to the latest RE4HD release (v1.0.1: https://www.re4hd.com/?p=9471)
- [X] Data has been [validated with SFV](https://www.re4hd.com/?p=9454) (a guide to using QuickSFV is available at that link, use that with this 1.0.1 SFV: [BIO4-HDProject1.0.1-SFV.zip](https://github.com/nipkownix/re4_tweaks/files/8057899/BIO4-HDProject1.0.1-SFV.zip))

### Describe your issue here (drag+drop ZIP to attach it)

During certain cutscenes the game FPS drops down to the low 30s, one cutscene that this always happens with is the 9th Chapter 1 cutscene (bottom right): 
![image](https://user-images.githubusercontent.com/1755499/168458087-296c0345-584e-4170-93c8-63967e943b9b.png)

Seems the lag is caused by the `PenClothMove3` function, patching that to start with a single `ret` gets rid of the lag altogether.

It might be something to do with the `penClothAtCkParallel` calls inside there, that func can also be patched with a ret instead of the Move3 func, which will also get rid of the lag, so it's looking like that func could be the culprit.

`penClothAtCkParallel` seems to call SQRTF in a loop, if you hook that func and make it return 0, that also gets rid of the lag...
Tried replacing SQRTF with an SSE2 replacement as well, but didn't make any difference sadly..

emoose:
Added a fix for this into #185 - like I mentioned there I didn't notice any visual difference from this change, but there's always a chance it could have broken something - if anyone does notice any difference please let me know!

RiasatSalminSami:
Does it only affect this cutscene or the whole game? I assume this does something with the cloth physics?

linkthehylian:
> Does it only affect this cutscene or the whole game? I assume this does something with the cloth physics?

It's only this specific cutscene afaik. It was immediately noticeable when testing the subtitles in https://github.com/emoose/re4-research/issues/11#issuecomment-1126848598.

emoose:
The patch in #185 changes a part of the Em30 / Em3f character cloth init funcs, so it _might_ have some effect in other cutscenes besides the one that we know has issues, if any other cutscenes make use of those two chars anyway.

[...]

PR #192, opened 2022-05-19
EnableFastMath: reimplement VEC/MTX functions to improve FPS

emoose:
More info in #188, this reimplements the `penClothAtCkParallel` function which caused framerate issues when using HD Project, somehow the reimplemented version seems to run at 60+FPS pretty much with no changes needed.

Added this to `FRAME RATE` section since this is mostly for boosting FPS, can always be moved somewhere else though.

Tried un-decompiling the func as much as I could, seems a bit more understandable now, but still not exactly sure what it's reading/writing though (seems to be distance related from all the squaring/square roots/etc, but not really that certain)

This includes a test-mode that can be enabled by building with `TEST_CKPARALLEL_REIMPL`, to allow comparing values between orig func & reimplemented one automatically, was very helpful to figure out what was going wrong with it.

..although by default our reimpl will always have slightly different values, since we compile with SSE/AVX which has less precision than x87 - fortunately you can disable it with /arch:IA32 - doesn't really seem like this precision has any affect in game.

I've only really used that test-mode in the cutscene mentioned at #188, but seems this parallel func might be used in a bunch more places in the game too, some things that seem to use it:

- Em2b - El Gigante
- Em31 - No. 0 Sadler (after transformation)
- Em35 - No. 1 Village head after transformation
- Em34 - No. 1 Village head before transformation
- Em18 - Merchant
- Em33 - InsectBossEvent
- Em30 - No. 0 Sadler
- Em3f - Sadler Ada Scenario Boss
- Leon jacket
- Leon holster
- Ashley skirt
- Ada dress
- Leon muffler(?)
- cObjBell
- Wesker jacket

**A build of this PR can be found here:** [re4_tweaks-EnableReimplementedCloth.zip](https://github.com/nipkownix/re4_tweaks/files/8734700/re4_tweaks-EnableReimplementedCloth.zip)
If anyone is able to help test against the stuff in the list above it'd be really appreciated! Just need to make sure nothing goes really bad/explodes/etc, and that the cloth still behaves the same as the original.

You should be able to toggle the `EnableReimplementedCloth` option at runtime with it changing immediately, eg. in that Saddler cutscene you can pretty much control the framerate with it :)

Since this is a better fix I guess I can remove the saddler fix from my other PR - though the root NaN problem in HD Project which might have caused this lag isn't affected, tried finding where that was originated from but didn't have much luck, seems to be something inside `r140s00.evd` at least.

In future maybe we could add reimplemented PSVEC/PSMTX funcs under this option as well, since that was able to help gain around ~8FPS or so.

(PR tested against 1.0.6/1.0.6dbg/1.0.6jpn/1.1.0 too btw, all seems good)


emoose:
Huh, seems `SQRTF` might actually be the culprit, replacing that with a custom version (which runs 2x as fast) seems to also allow FPS to reach 60 in that cutscene, guess the original code was hammering it like crazy, not sure why my version didn't have the same issue though...

I'm looking into rewriting all the MTX/VEC funcs now, got most of them done, even just relying on VC2022's normal optimization usually makes the new functions about twice as fast 🥇 

Hopefully might help with most of the frame dips in the game, if anyone knows any more troublesome spots please let me know!

laymonage:
> Hopefully might help with most of the frame dips in the game, if anyone knows any more troublesome spots please let me know!

The intro cutscene in chapter 1-2 (where Leon and Luis wake up tied in the cabin) seems to have severe frame drops in the scene when Saddler injects the plaga to Leon, causing audio desync. Might be the same issue there, I haven't verified if this patch fixes it.

emoose:
> The intro cutscene in chapter 1-2 (where Leon and Luis wake up tied in the cabin) seems to have severe frame drops in the scene when Saddler injects the plaga to Leon, causing audio desync. Might be the same issue there, I haven't verified if this patch fixes it.

Yeah that's the one I've been testing with, the build in OP should hopefully fix that, but think that might only help with that one cutscene. The MTX/VEC stuff should help with even more slowdowns though, if there are any others at least.

Here's a build of the MTX/VEC fixes, seems to be working pretty well (haven't tested much gameplay yet though): [re4_tweaks-EnableFastMath.zip](https://github.com/nipkownix/re4_tweaks/files/8736039/re4_tweaks-EnableFastMath.zip)

Mister-Curious:
Ok so I was really excited to see if this would fix this one bug I have with some custom model that cause a lag spike when loading the room (the issue only happens when we are not initially facing the enemies, and then turn around to face them). Unfortunately this math patch didn't resolve the issue. Keep up the good work !

emoose:
Ah that's too bad, might still be something math related that slows it down though, game uses a ton of slow x87 code everywhere, this only patches a small bit that's shared between stuff, maybe we'll find where that's caused some other time.

The EXE you sent looks like it was stopping sprint toggle hook from being applied, did you patch something related to sprinting? IIRC we did post some patches for it on this github somewhere...

https://github.com/nipkownix/re4_tweaks/blob/6ffed6c91750e2338a029449efc85aac6e6d9c12/dllmain/KeyboardMouseTweaks.cpp#L432 is the part that's failing.

Here's a custom build of the FastMath stuff which skips the sprint patch, it'd probably be better to fix your EXE so it can work with all the re4_tweaks versions instead though: 
[re4_tweaks-EnableFastMath-v2-nosprintpatch.zip](https://github.com/nipkownix/re4_tweaks/files/8748629/re4_tweaks-EnableFastMath-v2-nosprintpatch.zip)

Mister-Curious:
@emoose, good work. I see now that I did hardcode you sprinting thing with:

```
83 E0 01 33 D2 0B C2 74 66
and change it to
83 E0 01 33 D2 0B C2 EB 66

Then search for
83 E0 40 33 D2 0B C2 74 17
and change to
83 E0 40 33 D2 0B C2 90 90
```
I cant remember why I hardcoded it, but if its an option that is tweakable in the ini I can just remove it from my binary edits (I should probably should do this anyway to avoid future issues).  Can you tell me if there is a ini setting for this spring thing?

Cheers

emoose:
There should be a `UseSprintToggle` in the INI somewhere that can toggle it for you, should be better than hardcoding it.

Err I dunno what that files doing there neither lol, sorry about that. If you're removing the hardcoded patch there's not really any need for that zip anyway, just use the -sse2 or any other re4_tweaks ver.

emoose:
Yeah that being faster is pretty strange, still not fully sure that it's working exactly the same as original really... I guess we could always remove that replacement now that the `SQRTF` reimpl seems to fix the original code fine, will have to check which one actually lets it run faster with the game running uncapped.

I only noticed yesterday that there's still a ton of PSMTX funcs I forgot about adding here, maybe could help a lot more than the VEC stuff since the game does seem to do a lot of matrix math, will look into adding those soon.

@Mister-Curious let me know about another lag issue caused by many enemies/models being loaded in on the same frame, seems the game only completely loads in the model once it's actually visible on screen, if there's a bunch of models that this happens to at once it seems to tank framerate/lag spike for a moment, but once it's finished the game seems to have no issue with the models afterwards.
(seems to be initing something DirectX related, maybe it only uploads textures to GPU once they're first on-screen or something)

Found a workaround by patching the func that chooses what models to render based on camera view, so it'll fully process all the models in the level, if we let that patch stay active for just 1 frame/tick after "loading..." text disappears & then disable the patch, that seems solves this first-time-on-screen lag, and hopefully shouldn't affect performance ingame since it'll be disabled after that. (if there is any lag-spike from this process players will just assume it's part of the loading screen too :P)

Looks like this helped fix the lag spike issues in his mod at least, of course this is only a workaround though, maybe worth trying to figure out what data is actually being inited/loaded by the code we're patching.
(It's something in the ModelRender function afaik, this patch is just making it so ModelRender gets called for all the models in the level, AddOtModelPosRadius is what checks that they're in camera view/radius first)

Not sure if forcing all models to be processed like this might cause issues if the level has a ton of them though, and not sure if that lag-spike issue ever happens in vanilla/HDProject neither, so maybe this fix would be better disabled by default.

nipkownix:
Great work indeed! I imagine this is ready to merge?
Also, about EnableReimplementedCloth: Maybe we should have that enabled by default? Unless I missed something of concern?

emoose:
Eh I still think we could probably remove `EnableReimplementedCloth` altogether now, since I'm still not sure why that works, while the `SQRTF` replacement (`EnableFastMath`) is a little more understandable (replacing the extremely slow vc2010 sqrtf call with a SSE equivalent), better the devil you know etc...

I was thinking that `EnableReimplementedCloth` is likely still faster than original-with-new-SQRTF (since reimpl would be full SSE, vs x87+SSE), so maybe would offer some benefit to some people, but really it's probably a miniscule difference, & only effect that one cutscene (and maybe the EmXX enemies listed in OP)

Not sure if that's worth us having a settings toggle for/possibly confusing users, I'd say maybe should just remove the reimpl. but leave it in commit history in case we want to come back to it.

nipkownix:
>  I'd say maybe should just remove the reimpl. but leave it in commit history in case we want to come back to it.

Sounds good to me. I didn't realize the difference was so small.

> Got one more thing I want to try adding to this PR (replace the vc2010 sqrtf itself with our own instead of just RE4's SQRTF func),

Cool! Let me know once everything is done. Thanks again!

emoose:
@Mister-Curious let me know about an EXE patch which stops working with `EnableFastMath` enabled, not good...

Had a look at the patch he sent, only thing that could be conflicting is that his patch works by hooking a call to `PSVECAdd`, but the way we hook is by changing the thunk itself, which his patch is still calling into, don't really get why it'd break...

Will try looking into it some more, maybe will just remove the PSVECAdd hook itself since it doesn't improve that much anyway, would still like to find the reason for things breaking though in case the other hooks end up breaking other patches too :/

Mister-Curious:
Thanks so much for looking into this. The **dinput8-fastMath-noVecAdd.zip** you sent me does fix the issue of the patch, but the lag spike thing is back. Was that something that was part of the **PSVECAdd** hook?

(apologies for double post, I sent a copy of this to your forum inbox as well). 

Cheers

emoose:
Sorry for the delay, cloth reimpl. has been removed now, should be fine to merge AFAIK.

The issue with PSVECAdd was solved by Mister-Curious, I think his patch was slightly changed to not affect PSVECAdd, still not really sure why it got broken though.

nipkownix:
Ah, good to know Mister-Curious' issue was solved. I'm glad we're still compatible with his .exe, even though he/his team made some heavy edits to the .exe itself.

>Sorry for the delay, cloth reimpl. has been removed now, should be fine to merge AFAIK. (did try adding a sqrtf replacement but it didn't really make much difference, meh)

Cool, no worries. I'll merge it and send Albert a build. I've been tinkering with a revamped installer that will also grab the latest version of re4_tweaks from GitHub and install it during the HD Project installation. We're probably very near the final v1.1 update.

Thanks!

Issue #199, opened 2022-06-07
Audio music & fx volume sliders (Sliders added!)


mrtree3:
### Describe the feature you'd like to see added

Is there any chance of extra audio options please?

Like a volume slider for music & fx that could be added to tweaks?

Love this work, many thanks.

linkthehylian:
Yes, please! This is such a good request and would negate having to adjust its volume manually via volume mixer.

emoose:
Hmm, will try looking into this soon, got some ideas for how to try adding it in.

Unfortunately the whole audio system the game uses is pretty undocumented, I tried finding as much from the GC/PS2 releases as I could, but they did also add a lot of custom stuff for the X360/Windows ver that uses XAudio2 which I'm not really sure about.

One very useful thing I just found out though, seems Viewtiful Joe mostly uses the same audio system, and we have some [pretty detailed symbols](https://github.com/emoose/re4-research/issues/3#issuecomment-1061447972) for that, one of those being this struct:

```
typedef struct %anon154 { /* size 16 */
  sint16 mas_bgm; /* bitsize 16, bitpos 0 */
  sint16 mas_se; /* bitsize 16, bitpos 16 */
  sint16 iss_bgm; /* bitsize 16, bitpos 32 */
  sint16 iss_se; /* bitsize 16, bitpos 48 */
  sint16 str_bgm; /* bitsize 16, bitpos 64 */
  sint16 str_se; /* bitsize 16, bitpos 80 */
  sint16 ssq_bgm; /* bitsize 16, bitpos 96 */
  sint16 ssq_se; /* bitsize 16, bitpos 112 */
} SND_SYS_VOL;
```

From what I can tell, RE4 seems to be using the same thing :)

mrtree3:
Thank you a lot for trying!

I'll have a go with the zip

Scoovy:
> Seem to have something working, if anyone wants to help test here's a build: [dinput8_volumeAdjustment.zip](https://github.com/nipkownix/re4_tweaks/files/8854964/dinput8_volumeAdjustment.zip)
> 
> BGM/SE volume can be adjusted in the re4_tweaks Misc menu, seems to mostly work fine, not totally sure if all sounds will be affected by it though - would be interested if anyone finds something that doesn't work with it.
> 
[...]

PR #200, opened 2022-06-07
AudioTweaks: allows seperate adjustment of BGM / SFX / cutscene volume

emoose:
As requested at #199, went ahead and added volume sliders to a new Audio section in the re4_tweaks UI too (who knows, maybe we'll figure out more audio fixes eventually 😄)

Most kinds of audio can have volume adjusted at runtime, ~~but it seems BGM usually needs to be reloaded first (eg by reloading the level), also added a note about that to the UI.~~

I'd guess the volume values we're adjusting here probably get copied somewhere when that audio starts playing, maybe there's some way to track down where & adjust the volume as it plays (looks like the values might be getting transformed a little first though, so might take some work...)

FMV cutscenes seem to use some Criware audio system instead of the stuff main game uses, but luckily wasn't too hard to adjust that as well.

Seems to work across all the usual EXE versions, haven't tested it that much in-game yet though, would appreciate hearing from anyone that tries it out! (build artifact available at https://github.com/nipkownix/re4_tweaks/actions/runs/2457817918)

Some earlier notes that might still apply:
>not totally sure if all sounds will be affected by it though - would be interested if anyone finds something that doesn't work with it.

>there's also 3 different values for both BGM/SE as well, not completely sure I'm changing all the correct ones, so if you notice some sound effect changing when you adjust BGM, or vice-versa, let me know

E: as noted in the issue linked above, looks like cutscenes don't have BGM/SE separated properly for some reason, all sounds end up being controlled by `str_se` value... maybe this is part of why they never added audio options themselves.
Looks like  BGM/SE both go through the same func to fetch the volume, but a flag check decides what kind it is, maybe there's some way to set up that flag on the cutscene BGM.

E2: looks like that flag originates inside `bio4str.hed` :/ Not sure if the format for that is documented anywhere.


emoose:
Added a workaround for the cutscene audio, seems `str_se` is mostly only ever used by cutscenes, so moved that over to be affected by BGM slider instead.

Only issue I've noticed is merchant dialog now gets affected by BGM slider too, I guess maybe that's handled same way as cutscenes, not sure if there's any way to change it...

emoose:
Seem to have volume adjustment during playback working, for the `str_*` sounds at least, this seems to let us adjust BGM & cutscene audio fine.

I think certain currently-playing effects still aren't updated though, but since effects don't play for that long it's not a huge issue IMO.

Build artifact at https://github.com/nipkownix/re4_tweaks/actions/runs/2463654847

emoose:
Should have the issue with BGMs fixed, now hopefully all BGMs should be adjustable during playback 😄 

Build artifact at https://github.com/nipkownix/re4_tweaks/actions/runs/2464507044

emoose:
Well there is one small thing I'm not really sure about yet, the `mas_bgm/mas_se` values, it looks like those are master values which the other values are then multiplied by, initially I updated those as well, but removed that so we could have finer control of the other values.

Right now though when the game changes those mas values itself (`SndSubScreenInit`), our `Snd_set_system_vol_Hook` will store the value the game is setting, and then scale that value by the users volume setting to find the new value for `str_se`/`iss_se`/etc.

If the game is scaling those str/iss values by `mas` already though, this would probably end up scaling it by `mas` twice - `SndSubScreenInit` sets `mas` to 1/2 of normal value, so our `UpdateVolume` then scales eg. `iss_bgm` by 1/2, and then the games volume calc code scales `iss_bgm` by `mas`, maybe ending up at 1/4 of volume?

I'm not totally sure how the mas scaling works though, when I first started playing with this it didn't seem like mas influenced the volume at all, but the code does show it's being added in somehow... 

Eh, I'll try having a closer look at it later on, maybe can just remove the code that scales by what game tried to set.

emoose:
Looks like the volume in subscreens was definitely lower with this hook at 100% compared to vanilla game, went ahead and removed the stuff to scale by value set by game, seems better now.

If game ever updated `iss/str` volumes itself then the code for storing that & scaling user volume level by it would probably be useful, but seems game only ever updates the `mas` volume, and since it seems game is already applying scaling for that itself already we don't really need to bother with it.

New build artifact at https://github.com/nipkownix/re4_tweaks/actions/runs/2475491191

emoose:
Should be good to merge now, haven't had any issues in the few times I've tested with it at least.

Only thing missing now is some way of changing volume of certain effects during playback (eg. the sound that game makes as it fades into main menu, not really sure how that sound is being played atm...), but like mentioned before those effects don't seem to play for that long, so not a big deal, but would be nice to eventually get them adjustable...

nipkownix:
Thanks!

PR #202, opened 2022-06-08
OverrideCostumes: write setting value to INI

emoose:
Small thing I noticed, seems OverrideCostumes setting is read from INI, but never got written back, don't think that was intentional, so here's a quick fix 😸


nipkownix:
Ah, thanks! Not sure how I missed that.

PR #205, opened 2022-06-10
Settings: read overrides from `re4_tweaks\setting_overrides` dir

emoose:
As mentioned at https://github.com/nipkownix/re4_tweaks/issues/5#issuecomment-1152738691

Added a txt file explaining the overrides into `EnglishSubtitleData` folder too, so should get copied into the build artifact zips.

E: btw in case you want to make a new release soon, here's a changelog I put together quick, hopefully not forgetting anything (set a couple of the bigger changes to bold too since not all of them are as interesting :p), feel free to use/edit it:

```
- Fixed incorrect window size when using WindowBorderless
- AllowAshleySuplex: restores Ashley suplex glitch back to game (previously only possible in the first GameCube NTSC release)
- Allow reading DAT/EVD inner-files from loose folders if they exist, instead of needing to repack into the DAT/EVD
- Patched game to allow enabling subtitles when using English language, if valid subtitle data exists
- **Added newly written English subtitle data to the game**, which was missing originally (thanks to linkthehylian, Zaphero1, nr1995 & laymonage for writing/contributing to these!)
- Fixed certain incorrect subtitles during radio conversations (thanks to linkthehylian for correcting them, and zatarita for helping to locate them!)
- EnableFastMath: replaces games older math functions with newer SSE-enabled equivalents, helps improve framerate issues in certain areas
- PrecacheModels: helps solve framerate drops when showing multiple new enemies on screen at once (disabled by default as vanilla/HDProject doesn't require it - mod authors may find it useful though!)
- **OverrideCostumes: allows changing character costumes at any point during gameplay**, and fixes missing physics & animation issues with certain costumes
- Updated SilenceArmoredAshley patch to allow setting changes on-the-fly
- **Allow tweaking volume levels of music/effects/cutscenes during gameplay**, via new Audio section in the re4_tweaks config menu.
- Now reads setting override INIs from re4_tweaks\setting_overrides\ folder, useful for mod authors that require certain re4_tweaks setting values (more details inside override_info.txt file)
```


emoose:
Maybe should move the `setting_overrides` skeleton / overrides_info.txt to a seperate dir instead, maybe something like `dist`?

Made a test at https://github.com/emoose/re4_tweaks/commits/master, let's see if CI can still merge it in the zip properly...

nipkownix:
Ah, I see you updated the DatTbl stuff to use rootPath too. Good call.

Everything looks pretty good.

Many thanks for the quick implementation, and for the changelog. I'll probably make a new release soon, since the HD Project 1.1's final release seems to be very close.

PR #206, opened 2022-06-11
AudioTweaks: fixup settings for INI overrides, & speedup build/startup

emoose:
Another PR for you :p mainly to make sure the new VolumeBGM/SE/etc settings work with override files too. (should have added this in the PR for it earlier, ah well)

To sweeten up this PR I also found a way to speedup debug startup time by a ton, just set some files that we aren't likely to try debugging/stepping-thru to use full optimization, now instead of needing to wait 10+ seconds after starting debug, game should start almost instantly when running under debugger, and AFAIK it shouldn't affect us trying to breakpoint any of our code :)

(E: changed it to only affect Patterns.cpp instead, seems that's the main file that affected startup speed)

E: also found a nice build-time speedup as well, release rebuild went from 30 seconds to ~6 for me now, nice 🐱

### DIFF
--- a/dist/Bin32/re4_tweaks/setting_overrides/overrides_info.txt
+++ b/dist/Bin32/re4_tweaks/setting_overrides/overrides_info.txt
@@ -5,5 +5,8 @@ This should be useful if you're creating a mod that depends on a certain re4_twe
 In that case you can ship a "Bin32\re4_tweaks\setting_overrides\MyCoolMod.ini" with your mod which includes the setting values you require.
 (recommend including the full folder structure with your mod, Bin32 -> re4_tweaks -> setting_overrides, only the folders themselves & your override INI are needed)
 
+Any settings in your override-INI will always override the users saved INI file, so please make sure to only include settings that your mod actually requires!
+(in other words, please try to keep your override-INI trimmed down to the minimum settings needed)
+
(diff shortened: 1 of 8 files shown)


emoose:
Huh, noticed an issue with the game applying DPI scale when in windowed mode, so if you eg. had 125% scaling, if you were using a 1920x1080 window it'd actually render at 2400x1350, with pretty noticeable blur from the scaling :/

Looks like we can just call `SetProcessDPIAware` to prevent it though, works fine on my 125% setup at least, would appreciate anyone that can help test that out!

nipkownix:
Ah, nice!

Good catch with the `SetProcessDPIAware` thing. The devs should just have.. done it themselves ._.
Classic QLOC moment right here.

I'll go ahead and merge it now, since I have some changes I want to push, and don't want to mess up your PR by doing so :p

I've added a Master volume control and changed the layout of the Audio tab a bit:
image

I've also changed the ranges from using `0.0f -> 1.0f` to use `0 -> 100` instead, since that's what games usually do.

Hopefully you won't mind, but do let me know if you prefer keeping it the way it was before.

emoose:
Sounds good, I wanted to change it to 0 -> 100 too but didn't get around to that, new layout looks nicer as well, thanks for taking a look at it!

PR #210, opened 2022-06-12
Settings: spawn new thread to handle writing out INI file, fixes #209

### LINKED ISSUE #209: Saving from cfgMenu causes lag/desync
Odd thing I just noticed, when you press save on the cfgMenu it seems to hang/stall for a second or so, if you press the Save button a couple times during a cutscene it seems to cause things to desync too :/

Not sure if it's always done this or if something we added recently started causing it, I guess it's probably something to do with the way DX9 is hooked for Imgui to render, maybe there's some way to write the config file from a different thread instead.

E: hm, seems lag goes away by updating `Settings::WriteSettings()` and making all the `iniReader.WriteFloat` etc have `useparser` as true, guess lag was probably caused by the `WritePrivateProfileStringA` func that it uses when `useparser` isn't set.

Unfortunately looks like that removes all comments from the INI though...

emoose:
Seems to fix the lag issue, added some mutexes to make sure only one thread runs at a time as well.

### DIFF
--- a/dllmain/Settings.cpp
+++ b/dllmain/Settings.cpp
@@ -1,6 +1,7 @@
 #include <algorithm>
 #include <filesystem>
 #include <iostream>
+#include <mutex>
 #include "stdafx.h"
 #include "dllmain.h"
 #include "ConsoleWnd.h"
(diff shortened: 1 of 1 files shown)


nipkownix:
Thanks!

PR #214, opened 2022-06-14
vcxproj: use %BIO4_BIN% environment variable as DLL dest path

emoose:
Really wish I'd thought of this sooner, don't currently have an E:\ drive on this PC so I've had to edit the project settings every time I've cloned the repo :p, with this we should be able to have it use our own paths automatically instead.

The env var can be added pretty easily, https://thegeekpage.com/environment-variables-in-windows-11/ mentions a pretty simple way (on Win11 you can pretty much just type "Env" into start menu and it should find it for you)

It'd be nice if we could fill the Debugging settings to point at bio4.exe automatically as well, but think that isn't saved in .vcxproj...

E: oh seems debugging stuff can be copied from the .user file into .vcxproj, sweet, added it below. (wouldn't work with % for some reason but luckily $ worked)

### DIFF
--- a/dllmain/dllmain.vcxproj
+++ b/dllmain/dllmain.vcxproj
@@ -61,7 +61,7 @@
       <ModuleDefinitionFile>..\Wrappers\wrapper.def</ModuleDefinitionFile>
     </Link>
     <PostBuildEvent>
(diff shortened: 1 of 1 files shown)


nipkownix:
Ah, nice!

> don't currently have an E:\ drive on this PC so I've had to edit the project settings every time I've cloned the repo :p

Sorry about that, my C:\ and D:\ drives are dying, so I had to use E:\ for Steam games :p

Issue #193, opened 2022-05-27
Quickturn Fix for Matilda


Scoovy:
### Describe the feature you'd like to see added

this is just a simple request to add the quickturn that you can do with other weapons to the matilda

Example:

https://user-images.githubusercontent.com/103059961/170780773-fff5f984-3456-4440-9966-eb9c5b89f4ab.mp4

emoose:
>User Qingsheng has shared on how to enable it in the exe

Do you know what version of EXE those are for? Tried checking in both 1.1.0 & 1.0.6 and looks like those addrs aren't valid in either (both are in the middle of an instruction...)

Scoovy:
> 

im not entirely sure but i know it works for 1.0.6 i just tried it and it worked here is the exe if you wanna check it out

https://drive.google.com/file/d/15Tn3hZLxMcJAva0tiIvsLoIqlHBtYOag/view?usp=sharing

emoose:
Ohh right they're file offsets, thought they were memory addrs lol, I'll try having another look at it later.

emoose:
Just had a look at those patches, they're pretty smart, qingsheng is on another level than me lol

Seems he patches `wep17_r3_ready00` to use `CameraControl::getCameraDirection` and store result inside this+0x7D8+0x30, and also patches `wep17_r3_ready10` to jump to `wep02_r3_ready00` instead, pretty neat way of doing it.

Will try looking into it more soon.

emoose:
Made a PR for it at #218, any chance you could check that build and make sure it works the same as the EXE patch did?

Build artifact can be found at https://github.com/nipkownix/re4_tweaks/actions/runs/2506930793

PR #218, opened 2022-06-16
AllowMatildaQuickturn: allows quickturning chara when wielding Matilda

emoose:
As requested at #193 

Figured out how qingsheng's patch worked, pretty neat method, wasn't too difficult to replicate it in the DLL.

Looks like Matilda can quickturn fine now, but note that this won't have an effect when using MouseTurning.

Tested with the usual EXEs, seems to work fine.
Many thanks to @qingsheng8848 for the patch & @Scoovy for letting us know about it!


nipkownix:
Cool. Scoovy hasn't answered you yet, but I believe everything looks fine, so I'll merge it.
Thank you!

Issue #44, opened 2022-01-24
`D3DCREATE_MULTITHREADED` flag removal


emoose:
Like mentioned at https://github.com/nipkownix/re4_tweaks/issues/5#issuecomment-954563654, the game uses the `D3DCREATE_MULTITHREADED` flag when creating the D3D9 device, which adds extra thread safety to the D3D funcs at a cost of performance.

Removing games frame cap showed removing this flag allowed going from 220FPS to 270FPS in one area, would probably be good for people that struggle running the game at 60 and end up with game slowdown due to it.

Unfortunately the game seems pretty unstable with this flag removed, obviously they didn't bother making it thread safe on Windows - however, it seems the game does have a pair of nullsubs around certain graphics-threading related things, kinda seems like they were meant to be a pair of funcs for locking/unlocking mutex, but that's just a guess.

(On X360 the `D3DCREATE_MULTITHREADED` flag is apparently non-functional, not sure if that means X360 always had thread-safety stuff added, or maybe X360 devs had to be more careful with threading, which would explain the nullsub pair - but doesn't explain why they removed the code inside ;_;)

---

So I did another one of my experiments, hooked the two nullsubs to call `lock()/unlock()` on a std::recursive_mutex - sadly this resulted in a crash on boot, seems the UnlockMutex hook was being called before LockMutex for some reason, maybe the code for locking it got removed or something, just got around that by adding a hack to skip first call to it.

Then the game just started hanging before intro movie, one thread was waiting on the mutex lock to get unlocked by another first, some reason there's a few LockMutex calls in the game without accompanying UnlockMutex after it - likely an accident, since the func for locking mutex also returns the D3D pointer, they probably added some Win32 specific code and needed a way to get D3D device, and didn't bother adding UnlockMutex afterward since Win32 didn't actually need it. (seems this missing UnlockMutex only happens around calls to `D3DXCreateTextureFromFileInMemoryEx` - guessing that func was probably added for Windows)

Hooking the calls to that D3DX func & making it use UnlockMutex afterwards seems to let it continue though, managed to load into a game fine with it, + haven't had any crashes yet (besides a crash on exit, probably not too hard to fix though)

(not sure how proper that fix is though, game seems to be doing something with a ptr returned from that D3DX func, so maybe it should only be unlocking after it's done with the ptr, instead of right after calling D3DX, not sure...)

[...]

Issue #62, opened 2022-02-02
The game froze once at the beginning (1.1.0, English, No Mods, New Game, Normal, Standart Costume)


NikkiTanaka:
The game froze when I tried to ‘check’ the broken bridge after the ambush at the beginning of the game. I could Alt+Tab, but I couldn’t actually switch windows or do anything at all. Even the mouse cursor disappeared. Alt+F4 didn’t work, too. So, I had to restart a PC.

The second time went perfectly fine. What’s interesting, during the first attempt there were no button prompts for actions (‘Check’, ‘Take’, etc.). After the restart of a PC, they appeared. I don’t know if this is connected to freezing.

And another thing: when the game froze, the sound was still on, but not from the cutscene with Leon looking down, but the ambient one.

Win10 (64-bit)
750 Ti
i5-3450 3.10Ghz
8,0 GB RAM
SSD
The game played on Steam

BIO4: 6,804 files; 37,4 GB (40 241 926 498 bytes [size]/40 254 124 032 bytes [size on disk])
Bin32: 7 files; 21,4 MB (22 498 912 bytes [size]/22 511 616 bytes [size on disk])

emoose:
Hm, do you use any non-English locale/keyboard layout on your PC? Since you mentioned `8,0 GB RAM` I thought I'd ask - game seems to do something using system locale settings, could be related somehow.

NikkiTanaka:
@emoose, yes, Russian layout.

emoose:
@NikkiTanaka sorry meant to get back to you sooner - about the button prompts, seems game had issue with Russian language not displaying any keys, that should be fixed in #54 now (will display English keys since game doesn't have russian charset), next re4_tweaks release will include it (you could download the build on that page too if you want)

For the freezing, maybe try checking file validity with the method at http://www.re4hd.com/?p=9454 - if any file is invalid it could be causing crashing/freezing.

NikkiTanaka:
@emoose, QuickSFV showed that everything is fine, no errors.

I'm not familiar with github and its etiquette, should I close the thread or will someone else do it when circumstances of the freezing will become clear?

PR #257, opened 2022-07-09
ReplaceFramelimiter: implements new reduced-CPU-usage limiter

emoose:
As discussed at https://github.com/nipkownix/re4_tweaks/issues/44#issuecomment-1178293241, this implements the `FPS_ACCURATE` framelimiter from https://github.com/ThirteenAG/d3d9-wrapper/blob/c1480b0c1b40e0ba7b55b8660bd67f911a967421/source/dllmain.cpp#L46

For me this gave a pretty nice CPU usage reduction:
>CPU usage on main menu went from 7% to 1%, ingame from 10% to 4%

[qingsheng8848](https://github.com/nipkownix/re4_tweaks/issues/44#issuecomment-1179459205) also mentioned:
>it works very well, my CPU usage is reduced from 40% to about 20%.

The code is pretty close to the old framelimiter really, but I think the Sleep(1) call helps reduce CPU usage of the busy loop, there's also no longer a 0.01666666... minimum frametime with this so variableframerate above 60 should work better (though still probably has many issues like the ones mentioned in #50)

This also adds two settings to INI [DEBUG] section, `DisableFramelimiting` which should be useful for performance comparisons, and `UseDynamicFrametime`, which makes it pass the actual elapsed frametime to the game, instead of fixed 60/30FPS frametimes - **_this should help with the slowdown issue when framerate drops_**, but not sure how well things will work with non-fixed frametime, likely will have a lot of issues, so it's disabled by default for now.

Using both `DisableFramelimiting` & `UseDynamicFrametime` might come in useful for finding some non-framerate-adjusted things in the game too, eg. I noticed aiming with laser-sight seems slower using mouse to aim when FPS is very high, so I guess that might mean the aim speed is slightly different between 30/60 too. (#258 should fix that)

Haven't added any GUI options for the `ReplaceFramelimiter` setting yet, not really sure how to add stuff without breaking everything... would appreciate if you could show how to add a UI option for it with the new layout 😅 

---

If anyone is willing to help test it out, a release build can be found at https://github.com/nipkownix/re4_tweaks/suites/7489502732/artifacts/307459942 - just be sure to set `ReplaceFramelimiter = true` in the dinput8.ini first!
(should work fine across all the usual EXEs)

We mainly just want to make sure the game still plays fine with this framelimiter in place, no idea if anything might be depending on the previous framelimiter behavior or not. If anyone finds anything broken please let us know!

### DIFF
--- a/dllmain/DisplayTweaks.cpp
+++ b/dllmain/DisplayTweaks.cpp
@@ -173,14 +173,17 @@ void Init_DisplayTweaks()
 {
 	if (pConfig->bReplaceFramelimiter)
 	{
-		// nop beginning of framelimiter code (sets up thread affinity to 1 core for some reason)
-		auto pattern = hook::pattern("6A 00 FF 15 ? ? ? ? 50 FF D7 8D 8D");
-		Nop(pattern.count(1).get(0).get<uint8_t>(0), 0x16); // 6549C3
+		// nop beginning of framelimiter code (sets up thread affinity to core 0)
+		auto pattern = hook::pattern("A3 ? ? ? ? 6A 00 FF 15 ? ? ? ? 50 FF");
+		uint8_t* framelimiterStart = pattern.count(1).get(0).get<uint8_t>(5);
+		pattern = hook::pattern("E8 ? ? ? ? 85 C0 75 ? D9 EE EB ?");
+		uint8_t* framelimiterEnd = pattern.count(1).get(0).get<uint8_t>(0);
+		Nop(framelimiterStart, framelimiterEnd - framelimiterStart); // 6549C3 to 6549D9 (1.1.0)
 
 		pattern = hook::pattern("B9 ? ? ? ? E8 ? ? ? ? 84 C0 74 ? E8 ? ? ? ? 83 F8 1E");
-		uint8_t* framelimiterStart = pattern.count(1).get(0).get<uint8_t>(0xA); // 654A1E
+		framelimiterStart = pattern.count(1).get(0).get<uint8_t>(0xA); // 654A1E
 		pattern = hook::pattern("8D 85 ? ? ? ? 50 FF D3 DF AD ? ? ? ? 8D 8D");
-		uint8_t* framelimiterEnd = pattern.count(1).get(0).get<uint8_t>(0); // 654B0A
+		framelimiterEnd = pattern.count(1).get(0).get<uint8_t>(0); // 654B0A
 
 		struct CallFramelimiter
 		{
--- a/dllmain/DisplayTweaks.cpp
+++ b/dllmain/DisplayTweaks.cpp
@@ -3,6 +3,7 @@
 #include "Patches.h"
 #include "Game.h"
 #include "Settings.h"
+#include <timeapi.h>
 
 int g_UserRefreshRate;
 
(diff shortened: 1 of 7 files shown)


nipkownix:
Wow, nice work!

> Haven't added any GUI options for the ReplaceFramelimiter setting yet, not really sure how to add stuff without breaking everything... would appreciate if you could show how to add a UI option for it with the new layout 😅

Ah.. sorry about that. I tried to make the new code easy to adjust, but I couldn't.

Basically, you have to change the table index ``(ImGui::TableSetColumnIndex(1/0))`` every time you create a new option. You also have to save the Y position of the last item in the current column before moving to the next, and restore it when you get back to that column.
I've just pushed a commit to this PR that shows what I'm talking about.
I'm sure thee must be a better way to do this, though. If you can think of a better solution, let me know :p

(I haven't implemented the other options myself because it's 2:30 AM here and I'm very sleepy lol)

emoose:
Oh nice, not as difficult as I thought heh, thanks for the explanation!

Had a little look at it and might have found a slight improvement, should allow us to add things without needing to track 0/1 across items etc, I'll add it to this PR in a sec.

Maybe possible to add the `ImGui::Dummy` into it as well, but not really sure since it looks like Dummy gets used after all the items are setup too, any idea if we could remove that final one?

nipkownix:
Amazing work with the helper functions! Thank you!

nipkownix:
Tested it a bit more here and everything went smooth. Couldn't reproduce the freeze you mentioned.

emoose:
I'm sure I heard about that freeze somewhere else before, can't find much about it now though.

Tried testing out `UseDynamicFrametime`, seems to cause the freeze at start of first level issue that we had way back, don't think we ever found the actual cause for that though unfortunately, maybe will try looking into it more soon.

nipkownix:
Ah, odd.
I'll see if I can dig something up tomorrow. Sorry for not helping much, I was a bit tied up reworking our ultrawide patches :p

nipkownix:
FYI, I just cherry-picked 798bb4744135fc15dc673ccccca063ea6a648743 into master so I can make a new release.

nipkownix:
Rebased this onto current ``master``.

nipkownix:
Been playing a bit more using this, and no issue so far. Still can't reproduce the freeze from before.
I _think_ we should merge this PR, and leave the new frame limiter enabled by default.

emoose:
Sounds good to me, I never had any freezing after that point I mentioned with it neither, hopefully was just some kind of fluke, or maybe just a vanilla bug.

Will have to try looking into the `UseDynamicFrametime` issue I mentioned above some more, would be really nice to get game actually playable with that turned on (rather than freezing on first cutscene...)

nipkownix:
> Will have to try looking into the UseDynamicFrametime issue I mentioned above some more soon, would be really nice to get game actually playable at non-30/60 framerates with that turned on (rather than freezing on first cutscene...)

Cool. Just rebased this again and turned the new frame limiter on by default. Hopefully we can get ``UseDynamicFrametime`` working at some point, but this should already be a big improvement to the game. Many thanks!

emoose:
btw @nipkownix about the freeze I had with the police car, seems that's been happening for a while, someone mentioned it on re4hd recently: https://www.re4hd.com/?page_id=9319#comment-203270
and seems we did get a report about it here too: #62, but that was when we had issues with fonts which reporter also had.
Wonder what could be causing that, maybe it's framerate related, since we did get similar freezes with cutscenes when changing framerate at #328 ...

PR #258, opened 2022-07-09
MouseDelta: apply deltaTime, makes aim speed consistent across framerates

emoose:
Small thing I noticed while testing #257, though even without that you can notice a difference between aim sensitivity at 30 vs 60, this seems to help make them more consistent.

Not sure if the non-rawinput `(ptrMouseDeltaX) = (int32_t)regs.eax;` path might also need deltatime applied too.. will try checking that soon.

E: seems like it might, will update PR in a sec.
~~E2: hmm, can't seem to adjust non-rawinput with the same method, dividing by deltaTime makes aiming go all over the place, even when deltaTime is 1? weird.~~
~~If anyone has any ideas please let me know, otherwise I guess this fix is just for rawinput atm...~~

E3: ah nvm, regs.eax was uint but needed to be int, seems to work fine now, pushed with new fix.

### DIFF
--- a/dllmain/KeyboardMouseTweaks.cpp
+++ b/dllmain/KeyboardMouseTweaks.cpp
@@ -180,10 +180,13 @@ void Init_KeyboardMouseTweaks()
 		{
 			void operator()(injector::reg_pack& regs)
 			{
+				double deltaX = 0;
 				if (pConfig->bUseRawMouseInput)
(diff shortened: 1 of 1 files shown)


nipkownix:
I remember doing something similar in d95b354a98ef74b, but we never ended up merging that PR :p
Thanks.

PR #266, opened 2022-07-12
Trainer: GUI for trainer related patches / enemy data editing

emoose:
As mentioned at https://github.com/nipkownix/re4_tweaks/issues/243#issuecomment-1181395554

Added a quick GUI for editing data of `cEm` instances, also added some trainer patches to it as well, currently includes an Invincibility patch & a 1-shot-kill weak enemies patch - both patches are linked up to debug flags so the tool menu flag editor should be able to enable them too (maybe in future we could add a way to bind certain flags to hotkeys too, or something like that)

E: also added a infinite ammo patch, thanks to @Mister-Curious for providing the bytes for it!

bio4_22-07-12_17-22-59

**This is currently a draft PR since we're still waiting on #257 to be tested**/merged, which this includes some of the GUI changes from, once that's merged I'll probably redo this branch again.

Only really tested with 1.1.0 so far, but everything is sig scanned etc, so hopefully should work on other versions too.

Also interested in any requests for things to add to this section, or any ideas on best way to save/load `cEm` changes, not really sure how to handle that atm.
(main issue is that there's no way to uniquely ID a cEm... afaik cEms can get loaded into any slot, so can't rely on the index of it to know where to load modded values into, also cEms can be created/destroyed any time during play, so the cEm being changed might not even exist when level is loaded)

PR #284, opened 2022-07-22
cfgMenu: add column helper functions

emoose:
Adds the column helper funcs from #266, probably better to merge this in separately since I might need to redo that PR anyway, and not really sure when I can get around to that (don't really know what happened with the commits there neither)

E: eh not sure why it got made as draft PR, think it should be fixed now.


nipkownix:
Cool, thanks.

>I might need to redo that PR anyway, and not really sure when I can get around to that (don't really know what happened with the commits there neither)

Something weird happened to ``master`` as well. I had to force push a fix, so maybe that's what ended up messing that PR.
Sorry, I just fixed it by rebasing. Should be good now.

PR #294, opened 2022-07-28
bug_report.yaml: update links for HDProject 1.1

emoose:
Included a link to the BIO4.sfv that came with 1.1, since I'm not sure whether that SFV is included in the update-only release.

E: ugh it got made as a draft PR again, should be fixed I think


nipkownix:
Thank you, I was about to do that as well :p

PR #298, opened 2022-07-29
LAApatch: prompt user before game window is created or any textures loaded

emoose:
Changed it to use standard `MessageBox` before game window is shown or any textures are loaded.

Had to move the actual call to our LAA check stuff to our `CreateWindow` hook instead, tried handling it inside `DllMain` but that just made `ShellExecute` hang instead of relaunching, AFAIK that's something to do with [loader lock issues](https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-best-practices)...

Seems to patch/relaunch it fine this way though, also tested with EXE renamed after launching to make sure error msg func still works, all seems OK.

Since the game window doesn't even have a chance to show before this I removed part of the text about needing to relaunch the game, could maybe even remove the messagebox after the patch has finished and just go straight to relaunching, idk though, maybe it's better to let people know that our patching worked :p

(really we could probably remove all the messageboxes and just do it pretty much in the background, and only prompt user if patching failed... not really sure about trying to edit game files without letting users know about it first though, guess it's probably best to give some kind of prompt)

E: force push just touched up one of the comments, and re-added the "Non-LAA executable detected!" log.

### DIFF
--- a/dllmain/EndSceneHook.cpp
+++ b/dllmain/EndSceneHook.cpp
@@ -280,14 +280,7 @@ void EndSceneHook::EndScene_hook(LPDIRECT3DDEVICE9 pDevice)
 	#ifdef VERBOSE
 	// Show the console if in verbose
 	con.ShowConsoleOutput();
-	#endif 
-
-	// Show the LAA window if needed
-	if (laa.LAA_State != LAADialogState::NotShowing)
-	{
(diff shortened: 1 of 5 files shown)


nipkownix:
Thanks for the quick fix.

Issue #303, opened 2022-08-02
Ability for ETS models scaling


Corazon-SA:
### Describe the feature you'd like to see added

Hi, First thank you very much for the amazing work you are doing hare
I am happy that now we have enemy scaling (thanks to you), and I am start to think about the ability of ETS models scaling
We already have the option to change the scale of ETS model, but it doesn't work .. Thank you

emoose:
I'm not really familiar with the different model types and stuff used by the game (and the code usually refers to stuff differently than what the filenames show...), could you give an example what ETS models get used for?

Corazon-SA:
> I'm not really familiar with the different model types and stuff used by the game (and the code usually refers to stuff differently than what the filenames show...), could you give an example what ETS models get used for?
> 
> E: ah looks like `EtcModelSet` might be what handles it, that makes use of an `ETS_DATA` struct to setup/init some `cEm` instances - looks like that struct does have a scale field, but haven't seen anything actually try reading from it, hm...

.ETS files contains the information about interactable objects, such as (breakable barrels, doors, breakable glasses, ladders .etc)
we can set the (position, rotation, scale) of an ETS objects, but unfortunately the scaling doesn't change anything. I believe scaling ETS objects can scale its hitbox too, which is awesome for new ideas
Thank you very much

emoose:
Hmm, any chance you can point me to where the ETS files are stored for me to look at? Guessing they're in some kind of container like BIN/EVD/etc but don't know which ones.

Corazon-SA:
> Hmm, any chance you can point me to where the ETS files are stored for me to look at? Guessing they're in some kind of container like BIN/EVD/etc but don't know which ones.
> 
> Seems like I can probably add code to update the actual cModel scale based on the data in the ETS (surprising they didn't add code for that already), but not sure if there's any ETS which already has some value set there which might mess stuff up, if so I guess would have to add some kind of flag like the ESL scale stuff.


 .ETS files are inside .UDAS files. .UDAS files holds multiple containers inside it.
 .UDAS files are located in st1,st2,stX... folders,, .UDAS files contain rooms data. 
Let me know if you need any extra information.
Thank you very much

emoose:
Ah thanks for the info, had a look at one ETS and seems there are some scale values set there already, ones I saw were all just set to 1 though, no idea if there's any that might be different... I guess we can just update it from the values there and see what happens, can always add a flag etc later on if it turns out that some do have invalid data there.

[...]

PR #305, opened 2022-08-03
EnableModExpansion: allows overriding scale/speed inside ESL entries

emoose:
Separated the ESL expansion stuff from #266, since I'm probably gonna redo that PR eventually, and I guess it's better to get this stuff in sooner than later.

This adds a `EnableModExpansion` setting which enables the hooks for allowing scale/speed to be set inside ESL, had to change how the flag for enabling speed/scale is set though, since it seems 0xA might actually be in use (I didn't see any code using it, but the field is defined)
Now instead of setting 0xA to 1, you'll need to set the first nibble of it to F instead.
eg. if 0xA was set to `03`, you'd set it to `F3`, or if 0xA is `00` just set it to `F0`.
With that set the last 4 bytes of the ESL entry will be treated as two uint16_t percentage values, first one as speed percent, second as scale percent.

Also added stuff to allow scaling in ETS as well, as requested at #303, more info in the comment below, seems to be working fine from what I've heard 😸 

Still need to fix the interaction issue mentioned below though... hopefully can figure something out for it soon.

### DIFF
--- a/dllmain/AspectRatioTweaks.cpp
+++ b/dllmain/AspectRatioTweaks.cpp
@@ -119,7 +119,11 @@ void Init_AspectRatioTweaks()
 		void operator()(injector::reg_pack& regs)
 		{
 			// Code we replaced
-			__asm {fstp dword ptr[eax + 0x10C]}
+			uint32_t orig_eax = regs.eax;
+			__asm {
+				mov eax, [orig_eax]
+				fstp dword ptr[eax + 0x10C]
+			}
 
 			float* fCurHudPosX = (float*)(regs.eax - 0x1DCC);
 
--- a/dllmain/Game.cpp
+++ b/dllmain/Game.cpp
@@ -116,6 +116,17 @@ uint8_t* GameSavePtr()
 	return *g_GameSave_BufPtr;
 }
 
+bool IsGanado(int id) // same as games IsGanado func
+{
+	if (id == 0x4B || id == 0x4E)
+		return 0;
+	if ((unsigned int)(id - 0x40) <= 0xF)
+		return 1;
+	if (id < 0x10)
+		return 0;
+	return id <= 0x20;
+}
+
 // Original game funcs
 bool(__cdecl* game_KeyOnCheck_0)(KEY_BTN a1);
 void(__cdecl* game_C_MTXOrtho)(Mtx44 mtx, float PosY, float NegY, float NegX, float PosX, float Near, float Far);
--- a/dllmain/Game.h
+++ b/dllmain/Game.h
@@ -816,11 +816,31 @@ static_assert(sizeof(cEm) == 0x408, "sizeof(cEm)");
 struct ETS_DATA
 {
 	uint16_t EtcModelId_0;
-	uint16_t EtcModelNo_2;
-	Vec scale_4;
+
+	// normally uint16_t, but game only allows values 0x40 and below
+	// we patch game to read it as byte, which gives us a free byte to use as flag :)
+	union
+	{
+		uint16_t EtcModelNo_2; // vanilla
+		struct
+		{
+			uint8_t expansion_EtcModelNo_2;
+			uint8_t expansion_Flag_3;
+		};
+	};
+	union
+	{
+		Vec scale_4; // vanilla
+		struct
+		{
+			SVEC expansion_PercentScaleModel_4;
+			SVEC expansion_PercentScaleCollision_A;
+		};
+	};
 	Vec rotation_10;
 	Vec position_1C;
 };
+static_assert(sizeof(ETS_DATA) == 0x28, "sizeof(ETS_DATA)");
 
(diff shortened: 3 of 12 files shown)


FutonGama:
I tested the Speed and Size and its working nice like the old build. I need to test now the ETS size.

emoose:
Added code to also scale the ETS models collision radius/height as well, many thanks to @Corazon-SA for helping to test & refine it 👍 

We noticed that some objects have weird collision where the radius begins at an offsetted position inside the models, seems doors use this a lot - fortunately scaling that offset by the collision scale value _mostly_ seems to get collision to match up with the scale, but sometimes this scaled-offset isn't that perfect (maybe the original values weren't an exact match in the beginning, and now scaling them up makes that more apparent..)

To get around that I added a flag which lets you specify the model scale & collision scale separately, so modders can fine-tune it if they need to. 
The ETS_DATA struct was pretty much all occupied already, but it seems the `EtcModelNo_2` field (uint16_t) is actually restricted by the game to values 0x40 and below, so with a single patch we can make the game treat it as a byte instead, giving us an extra byte to use as a flag.

With the flag set the unused 12-byte scale area gets treated as two 6-byte vectors, one for model scale & one for collision scale, the vectors are 2-byte shorts for X Y Z, values are specified as percentages (so for 2x scale you'd set it as 200 / 200 / 200)
The flag is at 0x3 into the ETS struct, setting that to `F0` will enable this "short mode".

Some tools already treat the 12-byte scale area as a normal 12-byte vector, those can still be used with this fine, as long as that flag isn't set then both model & collision will get the values taken from there.

Getting late here and I'm pretty tired, I'll update OP with a better description tomorrow :)

FutonGama:
Its working, and the Collision Build too, ETS with collision working perfectly. Thank you. Darth is without computer so i'm testing it alone.

emoose:
Well looking into the checks for window "jump out" prompt to appear it seems `cPlayer::actionSelect` is responsible for the prompt itself, which calls a bunch of check funcs to see if you're near a window/door/etc.

`windowCheck` calls into `ChkWindow`, which seems to use `SceAtCheckFieldInfo` to find an AEV related to the window, that finally calls into `sceAtHitCheck` which seems to be what actually checks if you're near it, looks like AEV has `hit_dir_ang_48` / `hit_open_ang_49` fields that it uses to check against, I guess maybe modders can just update those ang fields if they need to, doesn't seem like there's much else we can change there at least...

Should be fine to merge, I'll add the debug build fix here in a sec, and maybe add some more info about what mod expansion does into settings_string.h too.

>Why is this, exactly? Anything broken there?
Hmm not really sure, haven't tried building it with the latest changes in a while, there are some commits in there like the column helper funcs that I'd guess might conflict though, not really sure if I can just remove those from the branch or not...

I've been thinking about redoing the Em manager side of it for a while though, would be nice if that could be 'popped-out' as a separate window so that you could see it while playing, kinda like the debug stuff added to ocarina of time here: https://youtu.be/r1BPWSADWck?t=206
(I like the way they added a toolbar menu at the top there too, if we ever added more debug-window kinda stuff it might be nice to add something like that too)

emoose:
>Should be fine to merge, I'll add the debug build fix here in a sec, and maybe add some more info about what mod expansion does into settings_string.h too.

Pushed, also checked with the usual EXEs, seems to be fine.

nipkownix:
> Hmm not really sure, haven't tried building it with the latest changes in a while, there are some commits in there like the column helper funcs that I'd guess might conflict though, not really sure if I can just remove those from the branch or not...

Ah, I rebased the branch some time ago, so I think there should be only minor conflicts? Haven't checked again since then.

> but it'd be nice if that could be 'popped-out' as a separate window so that you could see it while playing, kinda like the debug stuff added to ocarina of time here: https://youtu.be/r1BPWSADWck?t=203

Oh, that seems nice. We can definitely add an option to pop out the window. Not sure how we'd handle passing input to it, though. Unless you think the game shouldn't have input passed to it as long as the pop up is open?

> Pushed, also checked with the usual EXEs, seems to be fine.

Many thanks!

emoose:
> Unless you think the game shouldn't have input passed to it as long as the pop up is open?

I was thinking of it more like an overlay kinda, eg. you'd first open it from cfgMenu, be able to press buttons/select Ems/etc while still in the cfgMenu "mode" to choose what you want to monitor, but closing down cfgMenu would still leave the Em/debug window on screen, with inputs going to the game as normal.
Then if you need to adjust anything on that window (or move it/close it down) you can just bring up cfgMenu again, which would let inputs go to all the windows again.

Not sure how possible that is with the current setup, we do already have the console window drawing without cfgMenu needing to be shown at least, so I guess displaying it shouldn't be too hard, not sure about inputs though.

FutonGama:
Well if you guys need help to test something i'm here to do the necessary, every new features feels very good to me, loved. Count on me.

Issue #309, opened 2022-08-07
Game crash when select Bowgun


JackSound01:
### Validation

- [x] Game has been updated to the latest RE4HD release (v1.1: https://www.re4hd.com/?p=9552)
- [X] Data has been [validated with SFV](https://www.re4hd.com/?p=9454) (a guide to using QuickSFV is available at that link, use that with this 1.1 SFV: [BIO4-HDProject1.1-SFV.zip](https://github.com/emoose/re4-research/files/9207370/BIO4-HDProject1.1-SFV.zip))

### Describe your issue here (drag+drop ZIP to attach it)

In assignment ada when you select the bowgun, the game instantly crash.

nipkownix:
Could you please create a folder called "CrashDumps" inside Bin32, and then try to make the game crash again? A zip will be created with info that might help us find the cause of the crash.

JackSound01:
> Could you please create a folder called "CrashDumps" inside Bin32, and then try to make the game crash again? A zip will be created with info that might help us find the cause of the crash.

There is it https://www.mediafire.com/file/7s8wu9vvetcouxw/bio4.exe.20220807201204.zip/file

nipkownix:
It is crashing when reading stuff related to a model (probably the bowgun's model, as you mentioned). Doesn't seem to be caused by re4_tweaks. Seems to be something corrupted in your game. Are you using the HD Project? If not, you could try to verify the integrity of the game files on Steam.

JackSound01:
> It is crashing when reading stuff related to a model (probably the bowgun's model, as you mentioned). Doesn't seem to be caused by re4_tweaks. Seems to be something corrupted in your game. Are you using the HD Project? If not, you could try to verify the integrity of the game files on Steam.

I'm using the last version of Hd Project 1.1 and the last of tweaks.

emoose:
Definitely something related to the bowgun/wep50 model, could you try checking with the bio4.sfv to make sure your HDProject data is all valid? A guide is at https://www.re4hd.com/?p=9454 ("The game won’t open or there are some fatal crashes (I)" section)

Besides that I did notice you're playing at 3840x2160, wonder if that could be causing some issue with memory maybe, could be worth testing if a lower resolution stops the crash at all.

JackSound01:
[...]

PR #310, opened 2022-08-08
SDK: move game structs to new dir, seperated by known filenames

emoose:
Hey, wanted to get your thoughts on this, the Game.h seemed to be filling up (and will probably get even bigger soon :p), so figured we should probably split it up.

Tried separating things here by filenames I've seen in the symbols, 90% of these are actual header names the game used, but some I only saw as .cpp/.c and figured there was probably a .h for it too.

Do you think separating like this is fine, or is it maybe a bit too far? We aren't a decompilation project or anything like that so maybe naming stuff like this is a bit useless... there are a couple files in here which pretty much just have a single struct inside too, maybe more could be added later too, but not sure.

It might be convenient though since the symbols can be split into .cpp files pretty easily, so can pretty much just use the same filename that a struct was found inside (the symbols don't actually say which header a struct belongs to though, mainly just what .cpp/.c/.obj some code was from, so have to guess it from that... should be close enough though)

Also updated some of the structs with the new findings, need to make sure that things still match up/work as we expect before merging though (mostly seems to be fine from what I checked, but wouldn't be surprised if some offset got broken somewhere...)


nipkownix:
> Do you think separating like this is fine, or is it maybe a bit too far?

Nah, this is fine, actually. I was trying to think of ways of splitting Game.h somehow, but your solution is better than anything I came up with. Thanks for that!

> Also updated some of the structs with the new findings, need to make sure that things still match up/work as we expect before merging though (mostly seems to be fine from what I checked, but wouldn't be surprised if some offset got broken somewhere...)

Ah, cool. I'll do some testing later, then.

nipkownix:
Everything seems to be working just fine. I'll go ahead and merge this, so I can push a fix for #309 without creating conflicts with this PR.

If you have anything you want to change here, feel free to just push it to ``master`` directly, as I believe you have access to do that now :p

PR #325, opened 2022-08-25
PickupTransparency: fixes #324 by disabling on D3D device loss

### LINKED ISSUE #324: [Crash] Crash when alt-tabbing on item-examine screen, `IDirect3DDevice9::Reset` related
As reported by @albertre4HD at https://github.com/nipkownix/re4_tweaks/issues/320#issuecomment-1225976828 (some more comments about it there too)

>Hello!
>I think I've found a bug.
>When you take an item and I "Alt+TAB" the game while the item is rotating in the middle of the screen (game paused), the game crashes when I come back to the game. Here's the crash file:
>[bio4.exe.20220824184120.zip](https://github.com/nipkownix/re4_tweaks/files/9417954/bio4.exe.20220824184120.zip)

and tracked down by @nipkownix 
>Ah, if it only crashes when you Alt+Tab with full screen enabled, then it is probably triggered by IDirect3DDevice9::Reset. Might not be related to this hotkeys feature after all.

>Edit: Ah, lol, it is caused by RestorePickupTransparency. We finally figured out why QLOC disabled it: They couldn't/didn't bother fixing this crash. Pffff.

emoose:
Seems to fix the crash for me, like mentioned in #324 the transparency for the current screen gets lost on alt-tab, ah well, maybe 9Ex will help with that.

dxvk probably doesn't need this fix since it didn't have the crash issue, was thinking of adding something to detect dxvk so it wouldn't lose transparency, but hopefully it just won't send `D3DERR_DEVICELOST` in the first place. (E: yep seems it doesn't, nice)

If anyone wants to test it build is up at https://github.com/nipkownix/re4_tweaks/suites/7985352757/artifacts/341833441

### DIFF
--- a/dllmain/D3D9hook.cpp
+++ b/dllmain/D3D9hook.cpp
@@ -253,7 +253,27 @@ HRESULT hook_Direct3DDevice9::Reset(D3DPRESENT_PARAMETERS* pPresentationParamete
 
 HRESULT hook_Direct3DDevice9::Present(const RECT* pSourceRect, const RECT* pDestRect, HWND hDestWindowOverride, const RGNDATA* pDirtyRegion)
 {
-	return m_direct3DDevice9->Present(pSourceRect, pDestRect, hDestWindowOverride, pDirtyRegion);
+	static bool restorePickupTransparency = false;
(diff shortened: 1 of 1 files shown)


nipkownix:
Oh, totally missed this PR :p
Thanks for the quick fix. Nice thinking with checking ``D3DERR_DEVICELOST``, btw.

PR #330, opened 2022-09-06
ModExpansion: change flag offset 0xA to 0x0 - add 0x40 to set it

emoose:
This changes the byte we use to indicate that speed/scale should be updated, normally it'd update 0xA, but that field was already defined for use by the game (haven't seen any code that uses it though...), and also seems some stack-allocated `EM_LIST` instances could have random data there too.;

Instead of using that, it looks like the be_flag field has 3 bits unused (0x10/0x20/0x40) that we can appropriate instead, none of the UHD ESL files make use of them at least, and also couldn't find any code that use them neither.

So now to change speed/scale of Em, just add 0x40 to the first byte of the ESL entry (can usually just set first digit to 4)

With the flag moved that gives us 2 extra bits we can use to add further expansions (eg. the disable-laser-interaction flag suggested by @Mister-Curious, not sure how to implement it yet though)

This also fixes the issue with savegames/checkpoint not keeping the speed/scale reported by @Mister-Curious too, the flag should stay in place between savegames now, so speed/scale should still get updated.

Seems to fix it on my side at least, hoping to hear if it helps for others too, build can be grabbed from https://github.com/nipkownix/re4_tweaks/suites/8155427280/artifacts/354006147

### DIFF
--- a/dllmain/ModExpansion.cpp
+++ b/dllmain/ModExpansion.cpp
@@ -13,13 +13,8 @@ void SetEmListParamExtensions(cEm* em, EM_LIST* emData)
 	if (emData < pG->Em_list_5410 || emData >= &pG->Em_list_5410[256])
 		return;
 
-	// check emset_no_A field to see if custom speed/scale params have been set
-	// emset_no_A seems to mostly be unused, but the field is defined in the struct, so maybe something has it set
-	// hopefully nothing has all upper bits set though
-	if ((emData->emset_no_A & 0xF0) != 0xF0)
+	if ((emData->be_flag_0 & EM_BE_FLAG_MODEXP_SPEEDSCALE) != EM_BE_FLAG_MODEXP_SPEEDSCALE)
 		return;
-
-	emData->emset_no_A &= ~0xF0; // unset upper bits that we checked
 
 	if (!emData->percentageMotionSpeed_1C && !emData->percentageScale_1E)
 		return;
--- a/dllmain/SDK/em_set.h
+++ b/dllmain/SDK/em_set.h
@@ -5,6 +5,18 @@ enum EM_BE_FLAG : uint8_t
 {
 	EM_BE_FLAG_ALIVE = 0x1,
 	EM_BE_FLAG_SET = 0x2,
+
+	// if one of these two are set, game seems to swap between them when loading
+	// maybe as some way of telling if it's loaded from a save or from a file?
+	// only starts doing it if one of them is set though...
+	EM_BE_FLAG_UNK4 = 0x4,
+	EM_BE_FLAG_UNK8 = 0x8,
+
+	// flags for ModExpansion.cpp SetEmListParamExtensions
+	EM_BE_FLAG_MODEXP_10 = 0x10, // available for use
+	EM_BE_FLAG_MODEXP_20 = 0x20, // available for use
+	EM_BE_FLAG_MODEXP_SPEEDSCALE = 0x40, // indicates speed/scale values should be set from inside ESL data
+
 	EM_BE_FLAG_DIE = 0x80,
 };
 


FutonGama:
I have think in something. It is possible for example to choose if the enemy will spawn on first time you enter on the map or the second time for example? So, you will be able to make a map and if you back to the map diferent enemies will spawn.
Or, a priority to what enemy will spawn maybe using if is 0x10 or 0x20 will spawn after you re-enter on the map. 
I have thinked, i know the game uses the first byte of ESL for set the enemy alive or dead so 00 is "dead/disabled" and 01 is "alive/enable" but the game uses more than just 0 or 1 in the code. In the island for example in emleon06.esl you can see some enemies have 05 and 07.
Resident_Evil_4_ESL_Tool_K9k40bxic2
As you can see the game uses more than just 0 and 1. When we put 05 on a enemy they not spawn and 07 not spawns too. Its like put 00. Maybe with that can have a priority for what will spawn first or after. Or if you want maybe is possible to do that with the 0x10 or 0x20. And keep one for the disable-laser-interaction like Curious said.

emoose:
Well the flags used by game are `01 / 02 / 04 / 08 / 80`, so 05 would be `01 + 04`, and 07 would be `01 + 02 + 04`, I guess it's probably the 04 flag which stops it from spawning
Did see 04 being checked for in the game code but couldn't actually figure out the effect of it - you can see my notes for them all at
https://github.com/nipkownix/re4_tweaks/blob/2ade7b1b29d49d2f48506f4179be0207cc22a37c/dllmain/SDK/em_set.h#L4-L21

Maybe will have to look into that some more.

That's a neat idea to enable enemies between rooms, will have to check if the game is still processing the ESL entries for enemies in other rooms.

One thing that'd probably be possible is to have a flag that could enable/disable enemy after player loads save (or retries checkpoint), could probably scare someone with that :p

Mister-Curious:
@emoose thanks for this. About Futon's idea; we need to be careful though as a 0 flag doesn't always mean 'don't spawn', as it often means 'don't spawn until eventX happens', or 'don't spawn conditionx' is met.

FutonGama:
> 

Oh yeah, thanks for remember about that, i completely forgot that important information.

Mister-Curious:
> @Mister-Curious Did you have time to test this? No worries if not. Regardless, I think it should be safe to merge, unless @emoose believes otherwise.

No i havent seen Corazon online in quite some time, and he was the one who knows how to do this. We were testing things together at some point, but I don't remember how to use this function.

nipkownix:
Oh yeah, be nice to maybe include some documentation in the re4_tweaks folder about this. I haven't tried using any of this myself :/

emoose:
One thing to note is the issue Mister-Curious originally reported was that the ETS scale changes (that we added support for) weren't saved between retries/savegames, I got confused and thought he was talking about ESL scale/speed changes (another different thing that we also added support for :p), which also turned out to not be saving properly between retries/saves - this PR fixed the ESL issue, but the ETS problem is still there I think.

There's a lot of stuff added since 1.7.7.6 that need to be documented better really.. maybe will try writing up some kind of `for_modders.md` file soon explaining some of it (the ETS/ESL stuff, command-line params, best way of using the Em Manager/ESP stuff...)

FutonGama:
(the ETS/ESL stuff, command-line params, best way of using the Em Manager/ESP stuff, useful INI overrides...)

Em Manager its already saving? Because I'm really interested in editing things through Em Manager. I'm not sure if it will be possible to save one day. Otherwise it would be good to replicate the ESL tool within the tweaks and it would be very practical since we could have functions like copy and paste coordinates. We could select which ESL to edit and on top of that we could choose any .ESL file making it compatible with any additional side-loaded ESL such as Qingsheng's per map secondary ESL by selecting which file will be opened for editing. Idk, just maybe ideas out of the scope. Editing ETS using that is so cool too to positions. I loved.

nipkownix:
>One thing to note is the issue Mister-Curious originally reported was that the ETS scale changes (that we added support for) weren't saved between retries/savegames, I got confused and thought he was talking about ESL scale/speed changes (another different thing that we also added support for :p)

Ahh, I got confused too, then :p
Thanks!

PR #336, opened 2022-09-17
Fix r315 cabinet door position issue

emoose:
This fixes the issue with r315 cabinet door being positioned wrong after reloading the stage, thanks to albert for reporting this!

Since it's a very minor fix I figured there's no point adding an option for it, hopefully not much chance of anyone using EXE mods around that code neither, of course I've been wrong about that before though :P

Works across all the usual EXE versions etc.


nipkownix:
Heh, pretty interesting that they fixed it for SW, but not for the main game.
Thanks!

Issue #333, opened 2022-09-15
Porting & PC-specific issues


emoose:
Since the issues page has got a bit large and it's a little difficult to find what UHD issues should be worked on, I figured a separate page listing/linking to them would be useful.

This is by no means complete and mostly just which issues I can recall right now, feel free to mention any that are missing and I'll update this list.

This page mostly deals with three kinds of problems, issues that appeared during the porting process (GC/PS2->Wii->X360->PC), PC specific issues, and problems caused by 60FPS+ framerate.

(problems that have been fixed already should probably be mentioned here too, will update with those later on)

### Port issues

- Filter03 (glare) is broken, applies extra color to whole scene, should only be adding brightness to models that are close to light/fire sources - Filter03 code is changed a lot from GC version, but restoring GC code doesn't seem to help... (#5)
- Water texture is changed from GC, seems to have broken reflections in some spots (https://github.com/nipkownix/re4_tweaks/issues/175#issuecomment-1243043667)
- The saddler item pickup appears before end cutscene starts, making pickup glow appear throughout cutscene (#331)

### PC-specific issues
- Framerate is a lot worse compared to the older PC port, using `dxvk` can help improve it, but that could bring it's own issues too. Could be caused by overhead from the GameCube/Wii GX-API -> DX9 layer that it uses (dxvk wouldn't really change that though...), or maybe the D3D9 `MULTITHREADED` flag slowing things down? (#44, #255)
- Game-speed is tied to FPS, if framerate goes lower than the chosen FPS cap (30/60) then game will start to slow down - `UseDynamicFrametime` setting added by re4_tweaks can help with this, but also causes a lot of issues too, game isn't designed to work well with varying frametimes...

### 60FPS issues
- Water animation speed is doubled at 60FPS (#175)
- Animated textures also run at doubled speed (eg. flame effects)
- Certain animations (rifle reload) use a mode that locks the animation to 30FPS, other game anims are able to play at 60FPS+ fine, seems to maybe be disabling interpolation - re-enabling it brings a huge amount of issues though, likely something wrong with the animation data itself? (#23)
- Damage from enemies & weapons may be doubled at 60FPS

linkthehylian:
About the main menu background scrolling behavior, this actually still exists in the Wii edition and I've made sure to test it as well. Idk which port removed it first unfortunately, but it's definitely there on Wii.

No emulation was done to test this, btw. I still own my Wii and copy of the game for it haha.

[...]

PR #338, opened 2022-09-20
Misc: add fix for ShapeMove / PlSetFace single frame anim bug

emoose:
Should hopefully fix the bug with face animations not being shown, could possibly affect other anims that use `ShapeSet`/`ShapeMove` too, I think only hands & face anims use that (during gameplay and cutscenes), but could be wrong - in any case I'm pretty sure the original code was bugged and negative frame numbers weren't intended, so this fix shouldn't break anything at least, if we're lucky maybe it'd help fix some other things too :p

The posts around https://github.com/nipkownix/re4_tweaks/issues/333#issuecomment-1252207436 go into it some more (thanks to albert for testing & brainstorming it with me 😸), code comments should explain what was changed:

>ShapeMove has a bug that makes the current animation frame get set to -1, only letting it stay visible for a single frame
>(only happens with face anims that only have a single frame, which were meant to be shown until something disabled it)
>Seems to be due to the code at 6CCBDC in 1.1.0, on GC that would just set currentFrame to 0, but some reason UHD changed it to (currentFrame - 2)
>This code is only ran when currentFrame >= maxFrames, guess that was changed so once anim reached end it'd repeat the last 1 or 2 frames in the anim, while on GC it repeats from beginning frame 0 instead
>Bug happens when currentFrame is 1 and maxFrames is 1, that code will get ran, which results in currentFrame being set to -1...
>There is a flag check in that func that can force it to 0 instead, but that flag can affect some other things too, so it's better to hook it & check if below 0 instead

Seems to work fine across all the EXEs, might want to wait until it's been tested some more before merging though, not sure if it might affect cutscenes in some way.

E: since this only changes what happens at the end of the anim, and hand anims seem to use the same thing, I wonder if it'd have any effect with the weapon reloads - removing the 30fps flag from sniper reload still gave the jittery mess though, but https://github.com/nipkownix/re4_tweaks/issues/23#issuecomment-1187924132 mentioned removing the flag for crossbow would only break frames at end of the animation, maybe these changes could help with it somehow?

### DIFF
--- a/dllmain/Misc.cpp
+++ b/dllmain/Misc.cpp
@@ -1170,6 +1170,27 @@ void Init_Misc()
 		injector::WriteMemory(pattern.count(1).get(0).get<uint8_t>(1), uint8_t(0x65), true); // fadd -> fsub
 	}
 
+	// Fix face morphing (PlSetFace) during gameplay
+	// ShapeMove has a bug that makes the current animation frame get set to -1, only letting it stay visible for a single frame
+	// (only happens with face anims that only have a single frame, which were meant to be shown until something disabled it)
+	// Seems to be due to the code at 6CCBDC in 1.1.0, on GC that would just set currentFrame to 0, but some reason UHD changed it to (currentFrame - 2)
+	// This code is only ran when currentFrame >= maxFrames, guess that was changed so once anim reached end it'd repeat the last 1 or 2 frames in the anim, while on GC it repeats from beginning frame 0 instead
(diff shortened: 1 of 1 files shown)


nipkownix:
Oops, missed this one.
Pretty nice work! Wouldn't surprise me if the hermite stuff is broken in a similar way, honestly.

Judging by albert's test, I think we should be safe to merge this?

emoose:
Yeah played through first few stages and cutscenes all seemed fine for me too, should be good to merge.

Issue #364, opened 2022-11-14
Type III Controls: Keep Shooting During Context-Sensitive Attack


LethalPlacebo:
### Describe the feature you'd like to see added

When using the Type III controls that assigns shooting to Left Trigger + Right Trigger, Leon's shooting will be interrupted if he shoots an enemy, is very close to them, and causes them to enter a staggered state that opens them up to a context sensitive melee attack (ex: "press X to kick"). There is sort of a solution to this: if you release the right trigger and then press it again, Leon will be able to continue firing, but it would be nice if the problem didn't happen at all. The same thing happens with knife slashes.

Some background info that may be of use:

When using the Type I controls, where shooting is bound to RT + X, if you just keep holding X and shoot an enemy and the prompt comes up, Leon will enter the melee animation. The same thing happens with knife slashes. This reflects original GameCube behavior. Some form of this behavior could be affecting the Type III controls, where the game stops Leon from shooting, but doesn't transition to the melee animation since it's now bound to a different button?

This is why I labeled this as a feature request and not a bugfix, I do think the game is technically working correctly.

pas-de-2:
Weirdly, this doesn't happen with Type II controls, despite ostensibly having the same LT+RT binds.

LethalPlacebo:
> Weirdly, this doesn't happen with Type II controls, despite ostensibly having the same LT+RT binds.

Can confirm, with both gunshots and knife slashes you can keep attacking even if there is a kick prompts onscreen. Thanks for the info! So perhaps this behavior for the Type III controls is NOT intended after all. The fact that they Type II controls don't have this problem give me hope that this is something that can be fixed.

pas-de-2:
Found it. There's an explicit exemption made for type II config in `joyFireOn()`, they just forgot to check for type III as well.
```cpp
LABEL_12:
    if ( pSys->pad_type_B == 1 )
      return result;
    return 0;
```

nipkownix:
Nice job, @pas-de-2!

@LethalPlacebo, here's a test build: 
https://github.com/nipkownix/re4_tweaks/suites/9363574564/artifacts/441072190

LethalPlacebo:
Finally got home to test this, can confirm it works perfectly now. Can finally keep shooting or slashing with Type III controls, even while a kick prompt is onscreen. Thank you all! I will close this thread now.

PR #367, opened 2022-11-17
Fix Type III controls melee range firing bug

pas-de-2:
Fixes https://github.com/nipkownix/re4_tweaks/issues/364


nipkownix:
Nice work! Thanks! 
I can't test it right now because my controller isn't working ATM. 
Would you mind moving the hook to ``ControllerTweaks.cpp``, though? ``Misc.cpp`` is already a bit crowded.

pas-de-2:
Oh, sure thing.

Issue #183, opened 2022-05-07
[Feature request] Gamecube Balancing Restoration Option


Zaphero1:
### Describe the feature you'd like to see added

Quite a few exclusive upgrade values on gun stats were changed between the Gamecube release and every other release.
https://strategywiki.org/wiki/Resident_Evil_4/Weapons_and_upgrades
GC Red 9's exclusive upgrade increases its power to 5.0 instead of 6.5
GC Blacktail's exclusive upgrade increases its power to 3.4 instead of 4.5
GC Bolt-Action Rifle's exclusive upgrade increases its power to 18.0 instead of 30.0
GC Semi-automatic Rifle's exclusive upgrade decreases its firing speed to 0.4 instead of 0.8

There may be more balance changes than this. I don't know if anyone dislikes these changes that are mostly buffs, but I feel I should note that there are indeed changes there and I know some still stick with the Gamecube release.

LethalPlacebo:
This would be cool, but I think it's important to note something: only the North American version of RE4 for the GameCube had its values changed. As you pointed out, they were all lower, but to compensate the game gives you more ammo through out the game.

The original Japanese GameCube version had the stats for the weapons and item drops that are seen in every other port. 

To be clear I think this would be a cool addition to the mod, just important to note that when it comes to being faithful to the original GameCube release, what we have in the PC version now are actually the vanilla values.

sologuitarplaying:
> This would be cool, but I think it's important to note something: only the North American version of RE4 for the GameCube had its values changed. As you pointed out, they were all lower, but to compensate the game gives you more ammo through out the game.
> 
> The original Japanese GameCube version had the stats for the weapons and item drops that are seen in every other port.
> 
> To be clear I think this would be a cool addition to the mod, just important to note that when it comes to being faithful to the original GameCube release, what we have in the PC version now are actually the vanilla values.
The North American version is most likely the oldest, since it came out there first. 
I've read many claims of all sorts of changes that supposedly exist between the the NA version and subsequent ones; enemies behaving differently or being more or less numerous, ammo drops giving different amounts, etc. Wish there was more documentation of this stuff out there.

LethalPlacebo:
> 

[...]

PR #369, opened 2022-11-21
Added NTSC mode

pas-de-2:
Unlocks the NTSC exclusive difficulty adjustments for RE4 UHD.

Quoting my post in https://github.com/nipkownix/re4_tweaks/issues/183 as a summary of what this is supposed to do:

>- Normal mode and Separate Ways start at 5500 difficulty instead of 3500.
>- Assignment Ada is fixed to 6500 difficulty instead of 4500.
>- Easy difficulty is removed as an option from the title menu.
>- Shooting range requirements are 3000 per bottle cap instead of 1000.
>- The village stage in Mercenaries has a unique dynamic difficulty mode. In PAL the difficulty is fixed at 5500, but in NTSC it's unlocked, and you also gain 30 difficulty points every second. ***This requires a 60fps fix, because it's just calling GameAddPoint(LVADD_TIMECOUNT) every frame in R400Main(). At 60fps, you gain difficulty twice as fast!

TODO:
- [ ] the note in the shooting range still says that bottle cap requirements are 1000 for non-English languages

Not sure how to fix the shooting range note. I checked inside ss_file_01.MDT and found the 1,000 and 3,000 score requirement strings are already there, so the question becomes how to change the message ID of this particular object.


emoose:
Looks like `file_msg_tbl` defines the start MDT entry number for each file, `file_pages` also says how many pages/MDT entries are in each file (per language), seems it starts with the MDT entry number inside `file_msg_tbl` entry, and reads through that entry number up to the count inside `file_pages`.

`file_msg_tbl[0xE]` seems to be one for the target practice note, that sets start MDT entry as 0x3F/63, editing that to use the MDT entry num for the 3000 point msg (0x9D/157) seems to work, 0xC1BFF4 (byte) -> 0x9D in 1.1.0.

Probably wouldn't be too hard to pattern match that table, maybe should let it fail silently if it can't find the pattern though, in case some kind of EXE mod is installed.

nipkownix:
Nice job with this!
A question, though, since I haven't been following discussions related to this: I see you're manipulating checks that check if ``pSys->language_8`` is ``SYS_LANG_ENG_NTSC``, right? Have you tried just changing language_8 to ENG_NTSC instead? Just curious to know if the game works just fine with that changed, since I imagine you're doing it differently to allow these changes to work on different langs as well.

pas-de-2:
Yeah, the game seemed to run fine with language_8 = 1. Only issue I noticed is that the English subtitles patch doesn't work.

pas-de-2:
image

Managed to hunt down the texID of the START texture and swap it for NEW GAME 🥳

nipkownix:
With the clarified bottle cap requirements, I think we're safe to merge this?

pas-de-2:
Should be good to go 👍

emoose:
BTW just found a doc that goes through the changes between different regions/releases of the game pretty extensively, a lot of it seems like it's probably just dynamic difficulty related, but maybe there's some surprises there too: https://docs.google.com/document/d/1aXN-Kxt0jbt9HQk_YrUNEn4WalTyiw18u2XBeuhqMts/edit

pas-de-2:
Most of that tracks with what I've discovered. Main thing I'm trying to figure out right now is how they buffed the knife. Testing in the village on Professional, Ganados take the same number of handgun shots to kill in NTSC GC and UHD, but 5\~10 knife swings to kill in NTSC GC and only 4\~5 knife swings on UHD.

emoose:
Does changing UHD to 30FPS make any difference? IIRC most people assumed the increased damage was because of the framerate increase, hadn't really tried looking into it that much yet though. (I do know playing in 60FPS increases rate/accuracy of collision checks though, if that's related at all, which broke some speedrun strats iirc)

Those em** funcs are probably separated into the em** rel files in GC (which are stored inside the em**.drs files for some reason), got some IDBs for most of them, but could be some missing here: 
[re4_gcdebug_drs-rel-idbs.zip](https://github.com/nipkownix/re4_tweaks/files/10255002/re4_gcdebug_drs-rel-idbs.zip)

pas-de-2:
Nah, same thing at 30FPS. I know people report weird things with the knife doing multihits at 60fps, so I made sure that wasn't the issue. I also tested the GC PAL version in the same scenario, and that has the buff knife too, so it looks like an intentional balance change they made.

Thanks for the rel IDBBs! I'll take a look.

pas-de-2:
Aha, so it was in `GetWepDmVal`. First there is the `switch ( pEm->id_100 )`, which is where the game retrieves the damage modifiers for each weapon per em**. The Ganado modifiers are stored in `Dmg_tbl_em10`. The knife is wep_no 16, and the 16th entry in  `Dmg_tbl_em10` is 200. On NTSC Gamecube it's 100, so the knife was buffed to do double damage, and this carries over to almost every em type.

These table are interesting, because there's a lot of these changes. eg. the TMP also got a 40% damage buff vs Ganados. damage values for almost every weapon type were buffed by 33% vs em36 (Regenerator), 15% damage buff vs em39 (Krauser), etc...

Then I found this stuck in at the end of `GetWepDmVal`:
```cpp
  else if ( v3 == 16 )
  {
    Wep_7D8 = pPL->Wep_7D8; 
    if ( Wep_7D8 )
    {
      if ( Wep_7D8->field_42 <= 0 )
        return (int)((double)wep_nob * 1.3);
    }
  }
```
So it looks like the Wii port added a further 30% damage buff to the knife vs everything, that isn't active when using auto-knife, presumably to make manual knifing feel more impactful vs just using waggle vs everything.

The net result of this is the knife in UHD actually does 160% more damage to Ganados than it did on NTSC Gamecube.

Something else I noticed, and I remember someone mentioning this, the knife has a shorter range on NTSC Gamecube. Will need to find where those values are defined.

pas-de-2:
Last major thing I have to do is parse the differences in `RandomItemCk`. People have said that ammo drops more often in the NTSC GC version, and it's true. Right away I've noticed that the dynamic assist for low ammo kicks in more often, because the game is weighting shotgun and TMP ammo less when it calculates your inventory's total ammo value in `GetBulletPoints`.

It's interesting that `GetBulletPoints` is still checking for Chicago Typewriter ammo in the Gamecube version. Capcom clearly just gave up on this weapon at some point and turned it into a bonus gimmick. It'd be fun to try and mod in a balanced version one  day.

nipkownix:
@pas-de-2 Just noticed this seems to be crashing with the 1.0.6 debug .exe. Seems it is unable to find the pattern for ``file_msg_tbl_35``.

pas-de-2:
Huh, that's interesting. It looks like`file_msg_tbl[14]` is already set to 9F in the debug exe, so there's no `3F 00 01 00` entry for it to find. If you comment out that code and go to pick up the note, it says that scoring requirements are 3000, and if you set the language to something other than English, the game crashes when you try to pick it up.

That's not the only thing it doesn't like about my title menu edits... this exe is a pain in the ass.

nipkownix:
I mean, I'm not opposed to outright disabling this feature on 1.0.6 debug if supporting it turns out to be too much work..

pas-de-2:
It's no problem to keep supporting NTSC mode with this, but I might have problems with the Separate Ways difficulty menu feature in my other PR. Do we have a 1.06 debug IDB uploaded somewhere?

nipkownix:
I don't think we have much about 1.06 debug, actually. Unless @emoose has something locally?

pas-de-2:
Oh that wasn't too bad at all. I pushed fixes for everything to the Separate Ways PR.

emoose:
@pas-de-2 do you know if the NTSC/dynamic difficulty stuff affects QTEs at all? Someone on steam forums mentioned they might have been different in the original release:

>I was also just thinking that it would be interesting to have the option to disallow both dodge command inputs like in the original GCN -- do you happen to know if this option exists? (in other versions of the game, you can spam all four dodge input commands and it will succeed, but in the original GCN it would fail if you put in the wrong one first, so effectively you couldn't spam)

---

>I don't think we have much about 1.06 debug, actually. Unless @emoose has something locally?

Nah haven't really done much work on that other than naming stuff like SystemSave/Global, maybe should look into it more some time though, might still be some debug things there that we haven't reimplemented yet.

pas-de-2:
Hm, is that true? Playing the debug NA version in Dolphin, I can press all the wrong buttons, so long as I eventually press the right buttons before time is up. At least for the boulder after the farm, and the axe Ganado cutscene at the start of 1-3.

I remember having a similar thought before, that they must have nerfed the QTEs somewhere along the way, but maybe they were always like this?

Issue #332, opened 2022-09-11
Cursor Problems. 


FutonGama:
### Validation

- [X] Game has been updated to the latest RE4HD release (v1.1: https://www.re4hd.com/?p=9552)
- [X] Data has been [validated with SFV](https://www.re4hd.com/?p=9454) (a guide to using QuickSFV is available at that link, use that with this 1.1 SFV: [BIO4-HDProject1.1-SFV.zip](https://github.com/emoose/re4-research/files/9207370/BIO4-HDProject1.1-SFV.zip))

### Describe your issue here (drag+drop ZIP to attach it)

Everytime i open the RE4 Tweaks menu and close it the Cursor keeps on screen and i have to minimize the game to the cursor hide again. Even the hotkey to focus and unfocus don't hide the cursor. Another thing is the cursor not locked on screen and keep going to the other monitor.

nipkownix:
@FutonGama Could you check if this build improves something?

[dinput8.zip](https://github.com/nipkownix/re4_tweaks/files/10254769/dinput8.zip)

emoose:
Just had this happen after opening Globals window too, pressing F1 to close cfgMenu kept the Globals UI focused, with small cursor showing, pressing F5 then unfocused and gave it back to game, but made my system cursor start showing instead.
Tried opening pause menu/inventory/etc to hide it but nothing seemed to work, odd.
Was playing in windowed 1920x1080, and might have alt-tabbed after opening cfgMenu (but before opening Globals)

nipkownix:
Fixed a (unrelated?) focus issue I found when alt+tabbing with the UI open, but I couldn't reproduce the issue you mentioned :/

yaomtc:
Using Arch Linux, the game no longer allows me to aim and the cursor doesn't work in the F1 overlay. I can see a small cursor in the middle of the screen, and if I repeatedly press F1 I can see the cursor start to move but it's impossible to click and tends to go back to the middle of the screen. Happens on both X11 and Wayland with Plasma. I'll try installing GNOME and see what happens there.

EDIT: no difference. This is with the latest version of re4_tweaks. I haven't tried RE4HD or validating with SFV, but it was working when I last played (I downloaded it April 13 and played for a few days after that).  Must be related to a software update?

michaeljhkim:
[...]

PR #375, opened 2022-12-07
Trainer: add inventory item adding tool

emoose:
Got something working to handle adding to inventory for us 😸 
image

Instead of editing the inventory data ourselves and trying to handle placing it into free slots etc, this tries using same method game itself uses to add items, using the `PutInCase` func & setting up subscreen fields.

There's two ways it can work, can either try adding the item to case automatically & only display item-adding UI if there's not enough room, or it can always display the inventory UI for the user to place it themselves (the default, since it's more obvious that the adder is actually working with that)

Treasures/key items/other non-inventory-puzzle-items were a bit of a pain, can't use `PutInCase` with those since it'd crash due to no inventory model data for them, fortunately the games `itemInfo` func can tell us the type of the item, and then from that we can either call `PutInCase`, or use `cItemMgr::get` directly (which can handle the treasures etc for us)

There were a couple items marked as weapons which would still crash it too, not sure what's up with those, added a [list to filter those out for now](https://github.com/nipkownix/re4_tweaks/blob/inventory-adder/dllmain/Trainer.cpp#L2285) - no idea if there might be other items there that might cause issues though :/

Would appreciate anyone trying this out, tried adding checks to make sure items can only be added when game is in correct state (in-game with no menus showing), but I'm sure there's probably some way of breaking it though...
~~(eg. you can add item while on the pause screen, seems that's different to the other subscreen stuff, will need to find out how to detect when pause screen is active...)~~

Build can be found at https://github.com/nipkownix/re4_tweaks/suites/9722665738/artifacts/466312416, you can find the inventory adder in Trainer -> Debug Tools menu.

(E: build updated with the commits below)

TODO:
- [ ] Check whether all items can actually be added fine (also look for any treasures/key items that might have issues, like freezing when trying to examine them...)
- [x] Add check to prevent adding items on pause menu
- [x] Allow changing size of attache case (there are attache case items for each size, but doesn't seem to actually affect it, maybe need to update something in ItemMgr if we detect an attache case item is being added)
- [x] Save the "Open inventory after adding" checkbox state
- [ ] Cleanup treasure/key item names (maybe sort the list somehow too..)

Stretch goals (maybe not for this PR, but would be nice down the line)
- Some way of saving/loading inventory state to a file, so you could load your inventory between runs
- Editing items stats (a full inventory editor isn't that necessary IMO, but some way of adjusting the upgrades on a wep would be useful - maybe something that could work with the inventory mouse WIP?)


FutonGama:
I tested a few things.

Don't Worked:
Salazar_crest
Key_Sand_Clock
Treasure_Map_2
Treasure_Map_3

What i don't know:
If Time_Bonus work.
And i not sure what is Point_bonus.

Problems:
I added many treasures and the game "freezed" when i have tried to open treasure tab. (Happened one time)
Looks like some "File" note is not working, i noticed on the last ones from the list. (Need more tests)
When i tried to examine the Leon Jacket Bottle_cap my game crashed but maybe is a problem only here. Still, you can verify it.

Almost everything is working very good, i liked alot, amazing work!
And i believe need to see if you can add itens when on Hanningan Call and on Cutscenes, and if yes if this cause problems.

New Ideas:
Not sure if possible but be able to edit the Weapons atributes directly in to that. Capacity, Fire Power and maybe Fire Speed and Reload Speed. And starting items plus size of the Attach Case.
Ah and if you include that one day remember the native Handgun Firepower for some reason need to be Default or the others weapons firepowers will break, odd asf, i know.
Again, its just ideas. Thank you very much!

pas-de-2:
To get the tactical vest working, it looks like you have to call `PlSetCostume` and `PlChangeData`, eg. @0xD29651

Idea: add an item pickup `SndCall` based on the item's `type_2` from `itemInfo`. Looks like the values are in the switch in `sceAtGetItem`. Would give some nice feedback, especially when you don't have 'Open inventory after adding' checked.

emoose:
>To get the tactical vest working, it looks like you have to call PlSetCostume and PlChangeData, eg. @0xD29651

Oh cool, looks like that works fine - haven't tried with other charas yet though, guess we'd probably need to add some chara type checks around it for the charas that can't normally equip it.

The SndCall stuff is a nice idea too, will try looking into that soon.

nipkownix:
Finally had the chance to check this out. Very nice work once again, @emoose!

I was looking into adding filters by category, and the interface felt a bit too crowded. What do you think about moving this to its own page?

Got something like this in mind:
image

I imagine we should combine the categories with the same name, but I'm not sure if you still plan to rename them to something else.

Also: Are unk0 and unk4 just empty? Guess I'll remove the filters for those if so.

emoose:
Looks good, thanks for looking at it!

Yeah can probably combine the multiples, they do use separate values in the code but seemed better to rename them like that. Not sure what's up with Unk0/Unk4 neither, guess whatever used those was removed at some point. 

Only thing I'd suggest is maybe adding a "Select all" checkbox to select all the categories if ticked, and unselect when unchecked.

FutonGama:
@nipkownix Maybe get better if add Category like on Flag Editor for have less things on screen.
explorer_Ww666uiR2B

nipkownix:
> Yeah can probably combine the multiples

Ah, cool, done!

> Only thing I'd suggest for convenience is maybe adding a "Select all" checkbox to select all the categories if ticked, and unselect when unchecked.

Good idea. Just added it.

> Maybe get better if add Category like on Flag Editor for have less things on screen.

Hmm, but that would only allow us to choose one category at a time. The way it is right now allows us to have both weapons and ammo show at the same time, for example.

Combining the duplicates also got rid of an entire row, so I think we're good:
image

emoose:
>there are still some ways the game can crash. Adding the Bowgun as Leon, or the Striker as Ada, for example.

Would probably be best to prevent those from being added, maybe could just add some checks that set `bool disable` if it detects a bad chara/weapon combo, and add some text in the tooltip about it.

Might be interesting to find out what is actually making it crash there, IIRC there was some code that only allowed certain weps to be inited when a certain chara was being used, wonder if editing that code could help (guess there would probably still be missing anims for it tho)

nipkownix:
We could probably also let the user adjust the weapon upgrades before adding them to the inventory. Only weird thing is.. the game seems to store the upgrade state in ``pItem->weapon_6[0]``, but I have no idea how to manipulate that. Apparently this is what the merchant screen's code does when you upgrade a wep from there:

image

Gotta admit I don't really understand it, though.

nipkownix:
Hmm, I know in the OP you said we don't really need an inventory editor.. but I had to take a look anyway 😅

What do you think of doing something like this?

image

image

image

image

image

Also noticed something: ``WeaponId2MaxLevel`` doesn't take special upgrades into account. Special upgrades are a level above the max, so if a weapon's max capacity upgrade is 6, and there's also a special upgrade that increases the capacity, this special upgrade will increase the capacity val from 6 to 7. Can really do that with the trainer atm, will have to take another look at how to take special upgrades into account.

emoose:
Neat, that looks pretty good actually, the ones I talked about in OP were mainly the editors where you have to mess around with different slots, look for an empty one available to add stuff, then have a bunch of buttons & checkboxes everywhere for moving/copying things around the slots etc, seemed unnecessary to me when the game already has an interface to let you manage slots already IMO, all the editors I've seen like that just seemed bloated and difficult to use.

That's kinda why I went in the complete opposite direction with just a list + button to add stuff, the way you've done it there seems alright though, guessing it's just listing the contents without any worry about slot placement, and clicking shows that menu to change stack/delete/etc.

If the item adder is a separate window that hides after adding items that could maybe be a little annoying when trying to add a bunch of things at once though, maybe should only hide if the "open inventory after adding" is checked, or if PutInCase fails because of not enough room (so only when we open the item-add subscreen - could hide the whole UI when we show that actually, did mean to add that earlier on)

It's too bad that subscreen only allows a single item to be added at once, was thinking of adding some kind of queue thing too, but would probably seem janky if UI had to open/close for each single item :/

nipkownix:
> all the editors I've seen like that just seemed bloated and difficult to use.

Ah yeah, you're right. I never even had the patience to use Raz0r's inventory editor, for example. 
I went with a simple process, like you said. Just got the list of everything in the user's inventory and provided options based on which item type it is. 
I wanted to make a simplified inventory editor like this for a while, but I had no idea how I could add new items without having to make something really complicated. Your solution for how we could add items was a game changer IMO :p

> If the item adder is a separate window that hides after adding items that might be a little annoying when trying to add a bunch of things at once though

Oh, right, that's true. Changed it so you have to click "Close" to actually close the pop up:

image
(Added something to sort the items by type as well)

Ah, I've also made it so that the cfgMenu gets closed when ``InventoryItemAdd`` wants to show the inventory screen.

nipkownix:
@emoose Found an oddity with the punisher:
https://github.com/nipkownix/re4_tweaks/blob/fd2d025d1bcbb9f380c5ed76195d7e04f51fa33c/dllmain/Trainer.cpp#L765

Not really an issue, but we should investigate if Punisher id 64 even works anywhere in the game. (Could maaaaybe be the punisher you get as a gift from the Merchant?)

nipkownix:
> should maybe try looking into some way of freezing the game while cfgMenu is active...

Hm, yeah, you're right. I haven't really looked at that before. I'll see if I can dig something up soon. Been working on a save/load inventory option.

> Do you know where in the code it changes that ID to 64?

Haven't looked at that yet, but it would be cool if id 64 is a different wep xD

pas-de-2:
Small bug with the attache case upgrades: if you add an attache item, and then add a weapon afterwards, the size of the inventory doesn't update until you manually place an item in it, or open your inventory on your own.

nipkownix:
> Small bug with the attache case upgrades: if you add an attache item, and then add a weapon afterwards, the size of the inventory doesn't update until you manually place an item in it, or open your inventory on your own.

Should be fixed in b8d93e1.

> E: A 'Clear inventory' button would be nice as well,

Ah, yeah, should be simple to add.

emoose:
>Did directly calling InventoryItemAdd cause you any crashes before, though? If so, we can probably just come up with some type of queue system for this, I guess.

Eh not sure actually, think I just assumed it'd need to be called from in there to be safe 😅 
If all the stuff in `RequestInventoryAdd` works without needing to be in that hook I guess we can just move it out of there instead.

I did have an idea about all that RequestXYZ stuff though, could maybe make it use a `std::vector<std::function>` of callbacks to run in that hook instead, so you could schedule something to run there like:
```C++
Game_ScheduleInMainThread([item_id, item_count]()
{
  PutInCase(item_id, item_count);
  SubScreenWk->etc
});
```
which would add that lambda func to the vector, then the code inside `cSceSys__scheduler_Hook` would just go through and run each callback, and then clear the vector after.

Would make it much easier to run stuff inside that hook, rather than needing to add code to it / check XYZRequested bool / etc for all the things we want to run there... maybe will try looking into that later on.

nipkownix:
> Another suggestion: place a small [x] icon next to Search bar to clear the current search.

True, just added it. (also added it in the flag editor too).

> E: Some way to filter by weapons the player can currently use would be good, too.

This one is a bit trickier. I don't think the game itself knows which char can use which weapon...
We'd probably have to create our own list.

>I did have an idea about all that RequestXYZ stuff though, could maybe make it use a std::vector<std::function> of callbacks to run in that hook instead, so you could schedule something to run there like:

Ooh! That's a pretty neat solution, actually!

nipkownix:
> Hmm, since we need to be in-game and outside of menus for the item adder to work, should maybe try looking into some way of freezing the game while cfgMenu is active... I tried some things with stop flags before but always had issues with the in-game pause menu being able to overwrite them (or backing up the stop flags that we set to freeze it), usually ending with game being paused/frozen even outside of cfgMenu :/

Tried using the stop flags for this and I'm not able to reproduce the issues you mentioned with the in-game pause menu messing up with the flags.

This is what I've done:

```c++
bool ParseConfigMenuKeyCombo(std::string_view in_combo)
{
	if (in_combo.empty())
		return false;

	cfgMenuCombo.clear();
	cfgMenuCombo = re4t::cfg->ParseKeyCombo(in_combo);

	pInput->RegisterHotkey({ []() {
		bCfgMenuOpen = !bCfgMenuOpen;

		static int flagsBackup = 0;
		if (bCfgMenuOpen)
		{
			flagsBackup = GlobalPtr()->flags_STOP_0_170[0];
			GlobalPtr()->flags_STOP_0_170[0] = 0xBFFFFF7F; // Same flags the game sets during its pause screen
		}
		else
		{
			GlobalPtr()->flags_STOP_0_170[0] = flagsBackup;
		}
	}, &cfgMenuCombo });

	return cfgMenuCombo.size() > 0;
}
```

emoose:
Oh nice, seems to work fine here too, guess it was maybe just a problem with the way I tried implementing it lol

I was thinking maybe wrapping it with `Game_ScheduleInMainThread` might make it a bit more safe, so we can be sure main thread doesn't overwrite it at same time or something like that, but with that game seems to freeze again.

Seems that's because of `SPF_SCE` stop flag being set, looks like `gameMainLoop` checks that & stops `ScenarioMove` from being ran if set, which is what calls `cSceSys::scheduler`...

Something to keep in mind I guess, maybe changing our scheduler hook to hook into `gameMainLoop` instead could be worth looking into sometime too.

nipkownix:
Added the pause thing and tried to keep it optional. _Hopefully_ nothing breaks... but we should probably test it a bit more.

Couldn't reproduce the "you are dead" issue you mentioned, though.. Maybe I tried doing it wrong?

emoose:
>Couldn't reproduce the "you are dead" issue you mentioned, though.. Maybe I tried doing it wrong?

I noticed it only happened as that screen first appeared, think when the text is being zoomed in maybe, but once the Continue Yes/No buttons appeared it went back to normal, maybe you weren't opening cfgMenu early enough?

Or could be something to do with screen res maybe, I'll try checking some things with it later.

nipkownix:
Guess an easy workaround for all of that would be preventing our pause code from running if the player is dead? Not much reason to pause the game if that's the case anyway. Could maybe add a check for the player's + Ashley's hp.

laymonage:
> Guess an easy workaround for all of that would be preventing our pause code from running if the player is dead? Not much reason to pause the game if that's the case anyway. Could maybe add a check for the player's + Ashley's hp.

Might need a different way to check that, as Ashley can also be taken away (unless that also sets her HP to 0 somehow).

nipkownix:
> Might need a different way to check that, as Ashley can also be taken away (unless that also sets her HP to 0 somehow).

It does sets her HP to 0 once the animation of her entering the door begins. Guess that was an easy way for them to trigger the game over procedure.
Pushed some changes that disable pausing when Ashley or player are dead, and also disables it in menus.

nipkownix:
Rebased on top of current master. Looks like this branch was a bit behind (even behind the dxvk stuff), so rebasing was the only way. Hopefully everything is still working.

nipkownix:
I think we're probably safe to merge this? Unless someone encountered any weirdness with the inventory save/load feature...

emoose:
Think we're pretty close to merging it now, found a couple small issues with the save/load stuff though:

- Inventory saved fine, but loading it again seemed to put all the small items in first, without any room for the shotgun, making the subscreen UI appear for me to move stuff around for it - not sure if we could save/load item positions too? Would maybe need a different way to add it than `InventoryItemAdd` though I guess.. or maybe we could prioritise the larger items first somehow?

- Exiting that subscreen put me into the game frozen, was able to unfreeze by opening/closing cfgMenu luckily, I guess we could add something to InventoryItemAdd so it restores stop flags before opening subscreen.

- Trying to load an invalid json makes it throw an exception that we don't catch atm, bringing the game down

Not huge issues though, hopefully most of the remaining issues here are just small things like this, as long as it doesn't break the game by itself without user interacting with it I think we'd probably be fine to merge soon.

nipkownix:
> E: here's the json that did that in case you need to test with it:
[inv.zip](https://github.com/nipkownix/re4_tweaks/files/10270639/inv.zip)

Hmm, seems what is breaking this is our custom handle for attache cases inside InventoryItemAdd. Our json already tries to make sure the player's attache case will be able to fit everything, but during parsing of the json, the game is reaching a point where it tries to add a smaller attache case that's in the misc section of the json. That overwrites SubScreenWk->board_size_2AA, and everything else fails to fit properly.

emoose:
Ah that's weird, I guess when you upgrade attache case in game it must still be keeping the old size one in inventory too? Or maybe the save I'm using was modded somehow, this is just a random one someone uploaded here before IIRC.

We could check in `InventoryItemAdd` to make sure it's not changing it to a smaller one than the current size, but then players wouldn't be able to resize it smaller with the item adder :/
Maybe the JSON loader could set some IsImporting bool somewhere which would stop `InventoryItemAdd` from changing attache size, so only the first `AttacheCase` at the beginning would have any effect?

>If itemMgr::clear is doing the same, we could probably get rid of eraseAll...

Strange thing is that func doesn't seem to `erase` the items like our func does, looks like it just overwrites be_flag of each item, but then it runs some loop that overwrites `itemMgr->m_pAvailable_0` based on `itemMgr->m_flag_num_4`, no idea what that's for though.
Looks like it only really gets used on gameInit and inside r22c (shooting gallery), maybe it's just something specific for that map.

nipkownix:
Pushed something that should fix the inventory issue with that .json you uploaded.

>Ah that's weird, I guess when you upgrade attache case in game it must still be keeping the old size one in inventory too?

That could be it.. but I hadn't seen the issue with my saves. Maybe that inventory did come from a modded save after all...

emoose:
Seems to work fine now, thanks for the quick fix!

One other oddity I noticed is that the order of files changes after loading the json, looks like they're normally sorted in the order you collect them, but in the json we write out they get sorted alphabetically..

Looks like changing both save/load to use `using json = nlohmann::ordered_json;` lets it keep the order fine, didn't notice anything breaking from that at least.

I wonder if there's a way we could save what wep is equipped, since loading json will always unequip you atm...

emoose:
Tried adding equipped wep ID in the commit above, seems to handle saving/loading it fine, haven't tested too much yet though, but weapon hotkey stuff uses same method so hopefully will work okay.

Feel free to revert if you have a different method for it though.

DarkxVoid-loop:
> > Sim, provavelmente pode combinar os múltiplos
> 
> Ah, legal, pronto!
> 
> > A única coisa que propor por permitir é talvez adicionar uma caixa de seleção "Selecionar tudo" para selecionar todas as categorias, se marcadas, e desmarcar quando desmarcada.
> 
> Boa ideia. Acabei de adicionar.
> 
> > Talvez seja melhor adicionar Categoria como no Flag Editor para ter menos coisas na tela.
> 
> Hmm, mas isso só nos permitiria escolher uma categoria por vez. Do jeito que está agora nos permite ter armas e munições ao mesmo tempo, por exemplo.
> 
> Combinar as duplicatas também elimina uma linha inteira, então acho que estamos bem: imagem
> 
> @emoose, aliás, ainda existem algumas maneiras de o jogo travar. Adicionando o Bowgun como Leon, ou o Striker como Ada, por exemplo. Devemos tentar evitar que o jogador faça isso? Não tenho certeza de como faríamos isso.

@nipkownix Nipko Qingsheng modified Leon and Ashley's Pl to be able to use all weapons I don't know if Ada was changed too
It would be very good for you to add this already in the tweaks because we could play with any skin using all weapons 
The dll that makes this modification is HID.dll

nipkownix:
> Looks like changing both save/load to use using json = nlohmann::ordered_json; lets it keep the order fine, didn't notice anything breaking from that at least.

Ahh, nice thinking. I didn't realize it would be such a simple fix!

> Tried adding equipped wep ID in the commit above, seems to handle saving/loading it fine, haven't tested too much yet though, but weapon hotkey stuff uses same method so hopefully will work okay.

Tried a bit here too and it seems to work fine. Thanks!

> Nipko Qingsheng modified Leon and Ashley's Pl to be able to use all weapons I don't know if Ada was changed too
It would be very good for you to add this already in the tweaks because we could play with any skin using all weapons
The dll that makes this modification is HID.dll

@DarkxVoid-loop Oh, really? I did try looking into it before by analyzing what Raz0r's trainer does, but couldn't get it to work perfectly. Where can I find Qingsheng's dlll?

Also, unrelated, but you know FutonGama, right? Could you ask them to check if this fix worked? https://github.com/nipkownix/re4_tweaks/issues/332

DarkxVoid-loop:
@nipkownix Qingsheng's hand poses are perfect, I'll send you the dlls there on Discord 
I know him yes and I also suffer with this problem with the courses and for me it was still happening that the courses kept appearing

emoose:
Noticed a problem with Area Jump, maybe related to stop flags: if you sit at main menu and let movie start playing, area jumping will then take you back to menu for some reason, and menu is frozen and doesn't respond to inputs - if you wait a little while in this state it eventually does jump to the room you chose though, I guess movie is still playing in background and it only jumps once it ends? Not sure yet though, maybe just need to add something to `AreaJump` to make it stop any playing movie.

pas-de-2:
One more tool that would be nice is some kind of object inspector. I got an idea to programmatically disable or change certain item placements on r**init and it would be good to be able to see eg. their index value in RoomData.

emoose:
Was thinking of adding that for cObjs, could probably reuse the Em Manager stuff for it since things like flags/position/rotation/etc should be shared between them, maybe could just add a selector there that switches it between cEmMgr/cObjMgr and then hide/show some extra fields depending on the type.. maybe will try looking into that later on.

DarkxVoid-loop:
@emoose Would it be possible to add a merchant option in Trainer for us modders 
So we can change the value of jewels, weapons, upgrades and be able to modify the value of firepower, reload speed and capacity of each weapon?

pas-de-2:
> One more tool that would be nice is some kind of object inspector. I got an idea to programmatically disable or change certain item placements on r**init and it would be good to be able to see eg. their index value in RoomData.

To add on to this, the Em manager could use a `pEm->Item_id_3DE`/`pEm->Item_num_3E0` display, to see which Ems have pre-designated item drops.

nipkownix:
Since everything looks good here, I'll go ahead and merge it. Many thanks once again, @emoose.

pas-de-2:
I was testing some things with the handgun exclusive upgrade, and noticed that changing weapon upgrades with the inventory manager doesn't take effect on a currently equipped weapon until you unequip and re-equip it, or until you enter a load screen. The game needs to update the globals `weapon_lv_power_4FC3`, `weapon_lv_speed_4FC4`, `weapon_lv_reload_4FCA`, and `weapon_lv_blt_4FC5` that it uses to fetch your current weapon stats. Looks like these get set via `cItemMgr::arm`.

This fixed it for me:
```cpp
	itmPtr->setFirePower(customFirePower);
	itmPtr->setFiringSpeed(customFiringSpeed);
	itmPtr->setReloadSpeed(customReloadSpeed);
	itmPtr->setCapacity(customCapacity);
// update weapon power globals if this is the currently equipped weapon
	if (itmPtr == ItemMgr->m_pWep_C)
		ItemMgr->arm(itmPtr);
```

nipkownix:
Oops, I actually had the ``weapon_lv_xxx`` globals in mind, but totally forgot about them. Nice catch!

Issue #318, opened 2022-08-17
Title screen camera movement?


mrtree3:
### Describe the feature you'd like to see added

I've seen on another version of re4 you can move the camera at the title screen left/right slow or fast & also you can zoom in & out. I can't remember which port had these features though.

Would it be possible to add these in? I use a controller always but i think it was a keyboard feature. It would be nice to do these simple things with a controller. Best wishes!

LethalPlacebo:
This was already requested in issue #290, that feature actually isn't in this version of the game at all.

mrtree3:
Thanks for reply, it's not possible through tweals at all? That's a shame.

nipkownix:
It might be possible to replicate it, but I haven't looked at it yet. Not sure if emoose did either.

mrtree3:
> It might be possible to replicate it, but I haven't looked at it yet. Not sure if emoose did either.

Best of luck to you all

Issue #290, opened 2022-07-27
Restore Gamecube's main menu stick behavior 


Diegoeo9:
### Describe the feature you'd like to see added

In the original gamecube version (and Wii too?), there was this little behavior to the stick, that almost no one seems to know, (since it's removed from most versions), where if you move the right stick on the village main menu, up or down would make the background zoom in or out, and move it left or right you would change the direction the background is turning, the movement was similar to when you use the sniper scope. It's a little thing, but since you are restoring Gamecube features, i would like to see it brought back.
P.S. I can't find any examples of it on the internet, maybe only checking through emulator.

albertre4HD:
This is what he means:

https://youtu.be/795zZyxGXu0

Also, it would be great if the movie/trailer that plays when you wait a few seconds is active again (demo0eng.sfd and demo0.sfd) So I can remaster that video, too :P

linkthehylian:
> This is what he means:
> 
> https://youtu.be/795zZyxGXu0

Oh wow, I think I do remember this being a feature but never actually tried it out for myself. It'd be cool to see this be brought back!

emoose:
This might be handled by `titleLoop` function, on GC that seems to do a lot of stuff, and mentions a `zoom_in_limit.1654` variable as part of it, but on the PC release it just seems to have maybe 10 lines of code that scrolls the image around...

pas-de-2:
So this actually looks like it's pretty easy to restore 1:1, if I could just figure out how to hook and store the background texture ID_UNIT* ptr in titleLoop @0xC96D9B. I feel like I'm missing something obvious with my rusty C++, because I keep getting access violations.

I commented up the logic of the GC titleLoop function:
```cpp
unsigned __int8 *__fastcall titleLoop(int a1)
{
  unsigned __int8 *result; // r3
  double v3; // fp0
  double v4; // fp12
  double v5; // fp13
  double v6; // fp0
  double v7; // fp0
  double v8; // fp13
  double v9; // fp12
  int back_chain; // [sp+0h] [-20h]
  char v11; // [sp+8h] [-18h]

  *((float *)IDSystem::unitPtr(IdSys, 1, 0x28) + 0x22) = width_1653 * flt_80246BAC;// pos0_94.x = 0.0;
  *((float *)IDSystem::unitPtr(IdSys, 2, 0x28) + 0x22) = width_1653;// pos0_94.x = 900.0;
[...]

PR #382, opened 2022-12-10
Added option to restore Gamecube title menu control

pas-de-2:
For https://github.com/nipkownix/re4_tweaks/issues/290 and https://github.com/nipkownix/re4_tweaks/issues/318

https://user-images.githubusercontent.com/116680301/206851568-701b1ac3-4fa6-498a-967e-ee5b9e8f9fe0.mp4

Not sure if this was necessarily the best way to do this, but I wound up just writing a replacement titleLoop function.

In the GC version you zoomed in and out with the aim and knife keys. In the PS2/Wii versions, they mapped zoom to up/down right analog stick instead. I went with the PS2/Wii controls, because they're much more intuitive.

I also changed the RestoreDemoVideos key detection behavior to only trigger on active buttons, and not joysticks, like in the console versions.


emoose:
Awesome! Going with the Wii controls is probably better, seems to work fine with my Xbone controller - I never really played the GC/Wii versions much though so can't really say if it's working the same as those, at least it seems fine though.

Wonder if some kind of KB/M control could be added to it, maybe if mouse button is held it could follow the mouse pointer or something like that, AFAIK `PadRead` does have some code that sets the AnalogXX fields based on mouse movement, but that might only be ran under certain conditions (Del Lago fight?)

nipkownix:
Nice job!

About RestoreDemoVideos: I'm not sure if it is worth it to unset after SPF_KEY after calling KeyStop. SPF_KEY will be set to true once playback starts anyway, right?

pas-de-2:
When watching movies, the game stores all of your STOP flags in a `flags_STOP0_backup_180` inside `cSofdec::initWork`, and then restores them when it's over via `cSofdec::DispReset`. I had to track this down when I noticed the analog stick controls stopped working on the title menu when I quit out of a demo.

nipkownix:
Ah, so the flags are backed up, I see. Thanks for figuring this out!

I assume we can merge this, then?

pas-de-2:
Updated it to use `fAnalog` instead of `Analog`, feels much more like the console versions now without the big right stick deadzone.

I'm not familiar with how the KB/M inputs work. Are you thinking like holding right click to drag? Could do right click + mousewheel to zoom, then.

nipkownix:
>Updated it to use fAnalog instead of Analog, feels much more like the console versions now without the big right stick deadzone.

Should be careful using ``fAnalogRX/Y`` since, IIRC, they don't take any deadzone into consideration and will cause issues when the controller has even the sightliest amount of stick drift.
Seems I used 0.4 as a deadzone in ``LastUsedDeviceHook`` (ControllerTweaks.cpp), and that solved the issues Albert was having. Would probably be safer to use something like that here too.

> I guess Mouse3 to drag then?
I guess MMB could work fine. Or maybe something like holding down CTRL and moving the mouse.
I'm just not sure about the zoom. Hmm.

pas-de-2:
Yep, I borrowed code from `LastUsedDeviceHook`. I set the deadzone to 0.3. It's not completely gone, just noticeably smaller.

pas-de-2:
Going to leave this alone for now if someone wants to try their hand at implementing mouse controls. I took a glance at `PadRead`and my eyes started bugging.

nipkownix:
Cool, guess we'll merge this, then. I do plan on taking a look at implementing mouse controls, but not right now.
Thank you!

nipkownix:
Added KB/M support in ba593cff70e6d5552ef8373969b870bd99ccf0f5. Ended up using our own input funcs since PadRead is a disaster.. :p

PR #383, opened 2022-12-10
Trainer: improve "Disable Enemy Targeting" patch

emoose:
This improves the "Disable Enemy Targeting" trainer patch by adding some extra hooks for some AI funcs.

Two things are changed here, first I noticed `l_pl_378` was used in Em09 to check distance from player & decide whether to melee/walk/etc based on it, so added something to set that to a high number, seemed to help with a couple more enemies but most were still able to track player fine.
Some other Ems use `RouteCkToPos` & `RouteCkPosToPos` to check if it can pathfind to the player though, for those it seems we can just return false and that stops them from tracking properly.

(there were a couple other non-Em things that used those Route funcs too, but seemed like those could be detected & handled fine)

Need to test this out some more though, the few Ems I checked seemed to stop targeting player fine now, but wouldn't be surprised if some other Ems might not be affected by it..
Would appreciate anyone trying it out, please let me know if you find any Ems that are unaffected, or anything that seems broken with it!

Build can be found at https://github.com/nipkownix/re4_tweaks/suites/9788385305/artifacts/470741553 (patch can be enabled in Trainer -> Patches -> "Disable Enemy Targeting")

### DIFF
--- a/dllmain/Trainer.cpp
+++ b/dllmain/Trainer.cpp
@@ -505,6 +505,12 @@ void Trainer_Init()
 		patch_em10FindCk.SetPatch((uint8_t*)em10FindCk.as_int(), { 0x31, 0xC0, 0xC3 });
 		flagPatches.push_back(patch_em10FindCk);
 
+		// Em36 checks pPL->pos_94 directly inside em36AtkRtnCk, workaround by patching	it to immediately return
+		pattern = hook::pattern("89 45 ? F6 86 08 04 00 00 04 53 0F 84");
+		FlagPatch patch_em36AtkRtnCk(globals->flags_DEBUG_0_60, uint32_t(Flags_DEBUG::DBG_EM_NO_ATK));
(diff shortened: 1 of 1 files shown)


nipkownix:
Been testing this with some uncommon enemies I could think of. Everything seems to be working fine. Good job!
Only quirk I noticed is that Regenerators are still able to attack you if you get close to them.

emoose:
Added commit that hopefully stops regenerator from attacking now, seems that was checking `pPL->pos_94` directly inside one of its funcs, hopefully there's not too many other Ems that do it like that.

(seems `GetPlPos` is also exclusively used by Em code to fetch player position, could maybe add hook for that if any Ems still aren't affected)

nipkownix:
> Added commit that should stop regenerator from attacking now, seems that was checking pPL->pos_94 directly inside one of its funcs, cleanest way I saw to fix was adding a FlagPatch to make it ret, hopefully there's not too many other Ems that do it like that.

Ah nice, thanks for the quick fix!

Issue #331, opened 2022-09-10
Saddler Treasure appears before the cutscene ends


albertre4HD:
### Validation

- [X] Game has been updated to the latest RE4HD release (v1.1: https://www.re4hd.com/?p=9552)
- [X] Data has been [validated with SFV](https://www.re4hd.com/?p=9454) (a guide to using QuickSFV is available at that link, use that with this 1.1 SFV: [BIO4-HDProject1.1-SFV.zip](https://github.com/emoose/re4-research/files/9207370/BIO4-HDProject1.1-SFV.zip))

### Describe your issue here (drag+drop ZIP to attach it)

This is a bug the UHD game has since it was released. I tried to edit absolutely all stage data, but it remains there.

Saddler treasure/money should appear once the cutscene is over:
https://youtu.be/t9XsFR3dvWM?t=9961
![image](https://user-images.githubusercontent.com/82883088/189494489-80cc2423-8382-4e7c-abc8-b88be6570ad5.png)

But in the UHD (and later ports) it appears just when the cutscene starts, and you can see the blue halo it during several camera angles (one of them it covers the entire screen!!)
https://youtu.be/7J_kjTYKRmg?t=26846
![image](https://user-images.githubusercontent.com/82883088/189494529-0a7e2593-04eb-4d98-9ab2-8a64b1ee60af.png)

In the HD project, I made the halo smaller, but it's not a good idea because all the other bosses treasure halos are huge...

emoose:
Do you know which room number that map happens on?

albertre4HD:
Yep, r332 :)

emoose:
Had a quick look at this, seems `EmSetDropItem` is being ran twice for some reason: first it gets ran just after Saddler's death animation (causing the item pickup to appear during cutscene), but then it also gets ran after the cutscene ends too.

It being ran twice like this is probably unintended, since it looks like it only allows 1 item to drop from an Em anyway, so there wouldn't be any point in calling it twice.

Tried patching the call out before cutscene, and then restoring it while cutscene played (so that only the second call would have any effect), with that the item pickup was gone during cutscene & showed up fine afterward, so pretty sure this double-call is what causes it.

Seems like the first call right after death anim is probably the issue, unfortunately also looks like it's the same piece of code calling `EmSetDropItem` both times (at 0x4EE9A3), so not really something that can be easily patched out.

[...]

PR #384, opened 2022-12-10
Misc: add workaround fix for Saddler cutscene item glow issue

emoose:
From the comments:
>HACK: Fix for Saddler item drop appearing before cutscene begins, making the item glow show up through the cutscene
>Item gets spawned by em31_R1_Die_Normal when r_no_2 == 4
>Some reason r_no_2 gets set to 4 twice, once immediately after death anim, before the cutscene starts (causing the early item spawn)
>And then later R332RocketShootEnd calls into cEm31::setDieCancel, which forces r_no_2 to 4 (seems to be part of the cutscene code)
 
>We can patch out the first one by removing the r_no_2 increment call inside the r_no_2 == 3 block, which stops it from spawning early
>However GC (which doesn't have this issue) still seems to have the increment code, so this likely isn't fixing the root cause, but it's a workaround at least...
>Main issue could be the MotionMove check that r_no_2 == 3 runs before incrementing to 4 - maybe that only passed on GC once cutscene was over, but UHD passes immediately
>Need to check with GC debug and see if that also calls r_no_2 == 4 twice, and check what moment it actually increments it to 4...

More info at https://github.com/nipkownix/re4_tweaks/issues/331

Should be fine to merge for now, this only patches code related to Saddler, and albert mentioned that it seems to fix it fine - like comment mentions this is more a hack than a fix though, will probably look at GC debug some time to see if I can figure out why that acts differently (might take a while to figure that out though, so probably better to use this fix for now.)

Are there any places in the game where you fight Saddler besides the r332 ending though? This fix is sorta breaking the AI routines, relying on the cutscene after he dies to set them back to `r_no_2 = 4` instead, if there's somewhere you fight him where that cutscene doesn't play afterward it might cause some issue...

### DIFF
--- a/dllmain/Misc.cpp
+++ b/dllmain/Misc.cpp
@@ -1468,4 +1468,20 @@ void re4t::init::Misc()
 			}
 		}; injector::MakeInline<listRangeCheck_AllowSkipToBottomFix_LvUpMenu>(pattern.count(1).get(0).get<uint32_t>(0), pattern.count(1).get(0).get<uint32_t>(0x14));
 	}
+
+	// HACK: Fix for Saddler item drop appearing before cutscene begins, making the item glow show up through the cutscene
+	// Item gets spawned by em31_R1_Die_Normal when r_no_2 == 4
+	// Some reason r_no_2 gets set to 4 twice, once immediately after death anim, before the cutscene starts (causing the early item spawn)
+	// And then later R332RocketShootEnd calls into cEm31::setDieCancel, which forces r_no_2 to 4 (seems to be part of the cutscene code)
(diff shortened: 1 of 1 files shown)


nipkownix:
> Are there any places in the game where you fight Saddler besides the r332 ending though?

You do fight Saddler again in r51d (Separate Ways), but he's completely different there, so I imagine he uses separate code there. Outside of that, you don't really fight him anywhere else.

emoose:
Ah looks like r51d Saddler is Em3f, so shouldn't affect that hopefully, should be fine to merge then.

Issue #249, opened 2022-07-04
Matilda tweaks


NintendoFan:
### Describe the feature you'd like to see added

In the VR version of re4, the Matilda was changed to only complete a singular burst shot before stopping, meaning that if you hold the trigger down only 3 bullets will come out, you would need to repeatedly hold and release the trigger to get the same effect as in the original game that had been there by simply holding down the trigger. Is it possible to backport this change? I think this would make the gun a lot more likable for some players since it's way easier to save on ammo now.

Also, a way to make the Matilda switch to a normal fire mode would be nice, considering that there is even a visible switch on the gun to do this.

pas-de-2:
If you look back at the GameCube version of Matilda, this weapon makes a lot more sense. It fires slower, so you have way more control over how much ammo you're spending. It's probably why they didn't add a semi-auto mode. Here's the Matilda with GC frametimes:

https://user-images.githubusercontent.com/116680301/210195816-8e535720-62a5-4be0-ae65-3afdd2ea54d6.mp4

When they buffed it, they made the burst speed a lot more [like a real VP70M](https://youtu.be/RZ6yy1g1jyk?t=159) 👍 but they also increased the overall rate of fire for some reason, and turned it into an SMG:

https://user-images.githubusercontent.com/116680301/210195823-5eccf00f-a94f-45ea-bfee-b67579747222.mp4

If they had just buffed the burst speed, and left the second `MotionCheckCrossFrame` conditional in `wep17_r3_fire10` alone, it would have still been a more effective weapon, and we wouldn't need this LimitMatildaBurst 😕

https://user-images.githubusercontent.com/116680301/210195826-a3f0339d-477c-4da3-84e4-83ffbe140f59.mp4

NintendoFan:
They probably believed the Matilda was too weak in comparison to its more conventional counterparts, ie the Red9 or the Blacktail. As I recall all 3 of these weapons got buffed from the NTSC release to have more firepower, so that may be related? While the Matilda has the highest DPS out of any handgun, the merchant puts it best when he says "Although, it uses ammo rather quickly."

Also, is it possible that the 60 fps in this port compounds the Matilda speed to make it shoot faster than it should?

After looking at each individual video, I'd think I'd prefer the original implementation, yeah. But I've gotten used to this newer version already for years so I don't mind just using the new included option to limit bursts.

kazuya-555:
Maybe worth adding the edit from the 2nd video as an option perhaps.

PR #385, opened 2022-12-12
Added LimitMatildaBurst option

pas-de-2:
https://github.com/nipkownix/re4_tweaks/issues/249
I saw this suggestion and thought it was a good idea. This makes the Matilda feel a lot better when using type II/III config controls, as it's much harder to consistently control the rate of fire when using analog triggers. I suspect that's why RE4VR made this change as well.

(My dream would be if we could one day break the Matilda and its stock up into two items, so we could configure it RE2-style with semi-auto and burst modes)

~On second thought, maybe this option should just be something that forces `joyFireOn` to return false for every weapon, to be consistent. Let me know what you guys think.~
Edit: Hmm, I tried adding a universal semi-auto trigger for the other weapons, and it just doesn't feel good whenever you pull the trigger faster than the firing delay will allow you to shoot. It's fine with the Matilda though, since it has virtually no firing delay.

Edit2: Another option I've come up with is to only repeat fire when the right trigger is fully depressed. I still think I prefer this option, but that also feels way better than the default.

### DIFF
--- a/dllmain/Game.cpp
+++ b/dllmain/Game.cpp
@@ -84,6 +84,8 @@ namespace bio4 {
 
 	void(__cdecl* QuakeExec)(uint32_t No, uint32_t Delay, int Time, float Scale, uint32_t Axis);
   
+	bool(__cdecl* joyFireOn)();
+  
 	ID_UNIT* (__thiscall* IDSystem__unitPtr)(IDSystem* thisptr, uint8_t markNo, ID_CLASS classNo);
 };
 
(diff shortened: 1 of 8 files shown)


nipkownix:
Looks pretty neat. Now that I think about it, I never really used the Matilda before. This makes me want to give it a try, though.
Thanks!

NintendoFan:
> Looks pretty neat. Now that I think about it, I never really used the Matilda before. This makes me want to give it a try, though. Thanks!

It's a for fun weapon that if you use carefully, you can get more mileage out of it than the other handguns, but really if you didn't play resident evil 2 you're not going to see the real charm in this gun, since the Matilda is Leon's starting gun in that game. And that's mainly the reason why I like it a lot, because re2 is also a favorite of mine.

Issue #370, opened 2022-11-27
Force professional difficulty in SW/AA


Nitsug72:
### Describe the feature you'd like to see added

Would it be possible to force professional mode in separate ways/Asignment Ada?

pas-de-2:
Pretty easily done. Not so easy to add a difficulty select screen to the title menu like I wanted to do, though, it looks like all those textures get unloaded when `titleAda` and `titleSub` load their respective image packs. Could just make it a config option for now, like StartAdaPro or something and inline a hook at 0x63E301 in `gameInit`.

pas-de-2:
https://user-images.githubusercontent.com/116680301/208249312-b3123e2e-587e-4707-9138-0819219bf85f.mp4

I hacked something together for the Separate Ways menu. It's very much a hack though, and missing the fade transition that you get from calling `IDSystem::set` with an actual ID_FILE_HEADER. Test build:
[dinput8.zip](https://github.com/nipkownix/re4_tweaks/files/10252085/dinput8.zip)


Unfortunately, all the difficulty textures get unloaded and overwritten in the `titleSub` menu, so I couldn't do the same for Assignment Ada.

It looks like the data that defines the IDSystem effects and texture scale/positioning is stored in the .EFF and .UWF files of title.dat and ttl_ada.dat. Will need to figure out how that all stuff works to eventually do this properly.

Nitsug72:
That's amazing! <3
Is there a way to hide the debug console window?

pas-de-2:
Here's a more recent build without the debug console enabled:
[dinput8.zip](https://github.com/nipkownix/re4_tweaks/files/10298819/dinput8.zip)

Nitsug72:
Thanks! <3

PR #387, opened 2022-12-20
Added Professional mode select for Separate Ways

pas-de-2:
Fulfill's half of https://github.com/nipkownix/re4_tweaks/issues/370

I put this in a TitleMenu.cpp, figured we could start migrating some of the other title-related stuff from Misc.cpp there.


FutonGama:
Good work with that good stuff. Talking about that, do you think is possible to disable Normal difficult? Cause many players who play the mods choose normal but the modders always make the mod for play on professional. I Have enable the option to disable Easy so i imagined if can be possible disable normal too. Of Course with a box on tweeks to choose if want keep the dificulty on the title or not.

nipkownix:
Pretty nice work! Also, nice call with ``TitleMenu.cpp``.

nipkownix:
Are both ``isAssignmentAda`` and ``isMercenaries`` supposed to be checking ``SYS_OMAKE_ADA_GAME``?

pas-de-2:
Nope, lol, good catch.

nipkownix:
(Pushed a small menu-related feature here, since this is where we're moving things to TitleMenu.cpp)

Also tried looking at the Assignment Ada thing, but couldn't think of any decent way to do it.
Could _maybe_ just create our own textures to show there instead, but.. maybe it would be best to just add options in the .ini/cfgMenu to force difficulties for Separate Ways and Assignment Ada? That would at least make Professional Assignment Ada possible for now.

pas-de-2:
Nice tweak!

I wasn't *too* fussed about Assignment Ada, because it's sort of a mini-Professional mode already, in that it has no adaptive difficulty. But yeah some kind of .ini option could be good for those who really want it.

If we added `ForceProfessional` as a variable for modders, people could also just use that as a workaround for now?

nipkownix:
Pushed a new option to override the difficulty on any game mode. Should be enough for modders, I think. Doesn't affect existing saves, but I don't think that's a problem.

This got me thinking, though: Should ``OverrideCostumes`` and ``OverrideDifficulty`` be moved to the Trainer tab now? What do you think, @emoose?

Also, maybe ``SeparateWaysDifficultyMenu`` should be enabled by default.

emoose:
>Should OverrideCostumes and OverrideDifficulty be moved to the Trainer tab now? What do you think, @emoose?

Yeah that sounds good to me, I guess they could be added to Patches section, unless we want to make another Overrides tab for those? (could probably move speed/HP override stuff there too?)

Patches is starting to fill up quickly too, was thinking of merging some of the cheat-ish patches into a single column section, with the descriptions as tooltips, would free up some space but haven't tried it yet though, not sure if it might look janky or something lol.

pas-de-2:
I added checks to the `SkipLevelSelect` and `titleAda_levelMenu` structs to skip the difficulty menus whenever OverrideDifficulty is set.

Making `SeparateWaysDifficultyMenu` default sounds good to me 👍

nipkownix:
> Yeah that sounds good to me, I guess they could be added to Patches section, unless we want to make another Overrides tab for those? (could probably move speed/HP override stuff there too?)

Making an Overrides tab sounds nice. Will take a look at it soon.

Guess we should be safe to merge this. Many thanks for your work here, @pas-de-2!

PR #366, opened 2022-11-15
Refactoring

nipkownix:
@emoose Could you give me your opinion on this PR draft? This makes some changes I've been meaning to make for some time. Biggest change being the wrappers, but I'm also _trying_ to improve support for running from paths that contain UTF-8 chars.

While messing with the game on Linux, I noticed that the game crashes on start if the path it is running from contains special characters such as Japanese words. Before I could even dig into that issue, I realized something else: re4_tweaks itself was also having some trouble with special chars in paths.
I've changed some of our stuff to use ``wstring`` instead, but some other issues remain: I couldn't get our .ini reader to read files inside a path with special chars. There's also the ``looseFilePath`` from our sideload feature, which I haven't tested with special chars yet, but I imagine it wouldn't work as is.

Of course, none of this special/UTF-8 chars stuff is really important right now _unless_ we can make the game itself run from paths with those. (Do people even do that normally? Regardless, I imagine the game might also fail to load its own config.ini from the Documents path if the user's account name contains special chars..)

Do you think any of these things would be best left as is? The wrappers seem to be working nicely (from my tests). The ``wstring`` changes also aren't hurting, but they're pretty much useless right now, so no worries if we end up discarding this PR.

Ah, this should also fix the LNK4098 warning related to freetype. I noticed you pushed a workaround in 9bccd1ffbd3352c9f239cdc9cb6fde3bea981bf6 earlier. Turns out the problem was my fault for not including a debug version of the .lib as well. Sorry about that :p

PR #392, opened 2022-12-24
Created Gameplay tab in ImGui

pas-de-2:
For https://github.com/nipkownix/re4_tweaks/pull/366#issuecomment-1364454566

### DIFF
--- a/dllmain/AudioTweaks.cpp
+++ b/dllmain/AudioTweaks.cpp
@@ -127,4 +127,45 @@ void re4t::init::AudioTweaks()
 	// Hook SndCall call inside knife_r3_fire10 to allow restoring original GC knife sound effect
 	pattern = hook::pattern("83 C0 ? 50 6A 03 6A 01 E8");
 	InjectHook(pattern.count(1).get(0).get<uint32_t>(8), knife_r3_fire10_SndCall_Hook, PATCH_CALL);
+
+	// Silence armored Ashley
+	{
+		auto pattern = hook::pattern("83 C4 ? 80 BA ? ? ? ? 02 75 ? 83 FF ? 77 ? 0F B6 87 ? ? ? ? FF 24 85 ? ? ? ? BE");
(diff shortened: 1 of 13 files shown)


pas-de-2:
`UseSprintToggle` and `LimitMatildaBurst` are technically more like controller tweaks. I suppose we could duplicate `UseSprintToggle` in the Keyboard and Controller tabs, and `LimitMatildaBurst` in the Mouse and Controller tabs?

nipkownix:
Moved stuff to a new .cpp file and changes our .ini a bit. Should probably make sure everything still works correctly.. I've messed that up in the past ._.

> UseSprintToggle and LimitMatildaBurst are technically more like controller tweaks. I suppose we could duplicate UseSprintToggle in the Keyboard and Controller tabs, and LimitMatildaBurst in the Mouse and Controller tabs?

Hmm, I think since the implementation for them is the same for KB/M and controllers, leaving them in the Gameplay tab should be fine?

pas-de-2:
Oh, it's a shame that moving settings between tabs resets people's existing config settings. Maybe that's why you were opposed to new tabs?

nipkownix:
> Any reason the ini reader shouldn't just check for the whole file?

Hmm, just to keep things organized, I guess?
It isn't that big a deal IMO. But should probably add a warning in the release changelog mentioning that some settings may have reverted back to default...

pas-de-2:
What about grouping `DisableQTE` and `AutomaticMashingQTE` within the same column, like in https://github.com/nipkownix/re4_tweaks/pull/366#issuecomment-1364454566? I can't tell if you didn't like that or if you just merged over it with your existing changes, heh.

nipkownix:
Oh shoot, I think I may have accidentally overwritten that when changing the order of the settings. Sorry! I does look better.

nipkownix:
Everything looks good to me. Thanks again, @pas-de-2.

Issue #405, opened 2022-12-27
This application has requested the Runtime to terminate it in an unusual way.


InsideCat02:
### Validation

- [X] Game has been updated to the latest RE4HD release (v1.1: https://www.re4hd.com/?p=9552)
- [X] Data has been [validated with SFV](https://www.re4hd.com/?p=9454) (a guide to using QuickSFV is available at that link, use that with this 1.1 SFV: [BIO4-HDProject1.1-SFV.zip](https://github.com/emoose/re4-research/files/9207370/BIO4-HDProject1.1-SFV.zip))

### Describe your issue here (drag+drop ZIP to attach it)

Game Version - 1.1.0

I'm not really sure how to tackle this issue, and even though I think it might be on my end, in case it's important, I'm submitting this here in order to catch it. I just know it has direct relation to the Vulkan Renderer. Disabling it via the ini is what gets around the crash. Turning it back on results in it happening again.

![image](https://user-images.githubusercontent.com/111202797/209626208-2b21afa3-ef12-4dd3-aecf-075e61962754.png)

[I also have the crash via the CrashDump.](https://github.com/nipkownix/re4_tweaks/files/10306540/bio4.exe.20221227020405.zip)

nipkownix:
In dinput8.ini, under the line ``UseVulkanRenderer = true``, please add a new line with ``DXVK_LOG_LEVEL = debug``. This will create a ``dxvk.log`` in the ``re4_tweaks`` folder the next time you run the game. Hopefully there will be some helpful information there.

InsideCat02:
[Sorry for the delay! Here's the dxvk.log](https://github.com/nipkownix/re4_tweaks/files/10312124/dxvk.log)

nipkownix:
> Also, as a side note, a bug/problem I've noticed, setting the speed override hotkey to ' some reason makes the game refuse to launch. I was stuck figuring out why my game kept making refusing to load, and after a lot of repeat reboots and one by one redoing my settings, I narrowed it down to that.

Ah, you're right. I never noticed that before, but it seems to be a problem with the ini reader we're using:
https://github.com/ThirteenAG/IniReader/blob/85c2845b9dda0fa1f36612f8476d432dab1730d7/IniReader.h#L216

It is trying to remove single and double quotes from strings, but that obviously breaks our hotkeys. Hmm.

> Sorry for the delay! Here's the dxvk.log

Thanks, but it looks like there's nothing too helpful there...
[...]

PR #409, opened 2022-12-28
New iniReader implementation

nipkownix:
@emoose This should fix part of #405, as mentioned here: https://github.com/nipkownix/re4_tweaks/issues/405#issuecomment-1366625035

Figured it was time we made our own .ini reader to fit our needs instead of fighting the existing one.
~This is a bit simplified if compared to the original one, but I only focused on what we are actually using for now.~

Our extended ``CmdIniReader`` is now part of this too, so everything feels a bit cleaner.

Never tried doing this before, so I figured I should get your input before a potential merge. Everything seems to be working fine, though.

Edit: Was looking for a way to stop using ``WritePrivateProfileString`` since that is terribly slow, outdated, unsupported, and the only reason we had a separate thread to write settings. Came across [simpleini](https://github.com/brofield/simpleini), which can write back the .ini preserving comments and doesn't touch any of the outdated ``xxPrivateProfileString`` funcs. Seems pretty quick as well, so I removed the thread stuff we were using.

Should probably triple-check if everything is saving fine, though ._.


emoose:
Seems to work fine for me, settings + trainer stuff seemed to save fine at least, and cmd line args all looked okay, debug tool stuff like area jump still saved fine also.

I noticed this branch mentions "New refresh rate = 70" in console (I'm still using the OC'd monitor stuff I mentioned at https://github.com/nipkownix/re4_tweaks/pull/419#issuecomment-1384705616) - not sure if that means this is actually using that mode to display though, but IIRC I didn't see that message with the other branch.

nipkownix:
Ah, nice! Many thanks for taking a look.
I suppose there were no hiccups when saving settings? Guess we should be good to merge it.
Will most likely have to fix the other PRs though.. :p

PR #426, opened 2023-01-03
Added option to hide zoom control UI hints

pas-de-2:
Minor thing I wanted to add. Now you can make it look like GameCube.


nipkownix:
Looks pretty good, thanks!
I wasn't aware the GC version doesn't have these prompts. Always thought they were pretty unnecessary, guess the original devs thought so too :p

PR #428, opened 2023-01-04
Improved LimitMatildaBurst

pas-de-2:
Small problem with the old Matilda burst limit code was that if you pressed fire too early after a reload, nothing would happen. This fixes that.


nipkownix:
Good stuff. Many thanks!

Issue #436, opened 2023-01-09
Freecam improvements


emoose:
The freecam works pretty well, but could still use some improvements, right now it feels a little "floaty" for me, and also seems like the camera is still pivoting around something slightly in the distance, rather than being moved directly by our inputs.

I noticed setting the STATUS flag `STA_KLAUSER_TRANSFORM` seems to help with the pivoting issue a little, but the floatiness seems like it might be worse with that - found that flag being checked inside `CameraQuasiFPS::move`, think that func is what handles the pivoting for the camera, looks like there's a few different modes available in there, but only mode 0 and 2 ever get used? 

The func that calls that (`CameraControl::Move`) also calls into `CameraSmooth::move` & `CameraInterpolation::move`, maybe those are responsible for making it feel floaty too, will need to look into these some more.

emoose:
@nipkownix got some code setup to overwrite/restore `CamSmth.m_ratio_FC` when in freecam, could you give it a try with CE and see what you think of it first though?

It feels a bit better without the smoothing, but also kinda feels like distant things have slightly jerky movement, not sure what's up with that... maybe I'm just seeing things though lol

emoose:
Setting `g_transOfs[1].m_campos_0[0].x, g_transOfs[1].m_campos_0[0].z, g_transOfs[2].m_campos_0[0].x, g_transOfs[2].m_campos_0[0].z` to 0 seems to help get rid of the X axis pivoting:
`bio4.exe+7FB258+2C`
`bio4.exe+7FB258+2C+8`
`bio4.exe+7FB258+2C+2C`
`bio4.exe+7FB258+2C+2C+8`

Haven't had any luck with the Y axis yet though, setting those to 0 makes up/down movement on mouse act weird, and seems to break movement somehow (feels like movement is still moving camera where it would have pointed before m_campos_0[0].y were overwritten, maybe there's a different value it's using for it...)

linkthehylian:
I think the GC debug camera is what Boundary Break used for their most recent [video](https://www.youtube.com/watch?v=OezB3oGVEqo) as well. It also allowed them to freely move the camera outside of being loaded into a level.

I love exploring the outer parts of this game a lot so any improvements will definitely help.

linkthehylian:
Also @emoose, I noticed that when the free camera is enabled (using the build from https://github.com/nipkownix/re4_tweaks/pull/437) it immediately jolts down to Leon's boots. Just curious if that's intentional because it didn't do that before.

PR #437, opened 2023-01-09
Display: add `CameraSmoothing` setting to adjust smoothing ratio

emoose:
Seems to work fine, adds an INI/UI setting to Display section, and a `Game_SetCameraSmoothness` func that lets us adjust it in code (could let freecam code call that to override/restore it too)

Sadly setting the smoothness too low makes camera movements pretty jerky, as explained at https://github.com/nipkownix/re4_tweaks/issues/436 - would be nice if we could let smoothing only affect movements made by the game (eg. camera movements when player char climbs up/down), while player inputs would use a different value, seems game code treats them both the same though AFAIK :/

This should affect both mouse & controller movements, not sure if it might affect camera movements in cutscenes, I couldn't see any change in the starting car cutscene at least.

E: added code to disable the smoothing while in freecam, works pretty well there IMO, but feel free to revert if you like, or we could change it so it just reduces the smoothing instead of disabling it entirely (you can still change smoothing in cfgMenu when in freecam to try other values)

Build at https://github.com/nipkownix/re4_tweaks/suites/10270563416/artifacts/504249207 for anyone that wants to try it (smoothing setting is inside `Display` section)


emoose:
Finally found a way to get rid of the camera orbiting, this took me way too long to find out, was about to try decompiling GC's `debugCamera` stuff but decided to try out zeroing the `CAMERA_POINT` stuff first, and was surprised to see it actually worked :o

Had to adjust the `fFreeCam_depression = std::clamp(fFreeCam_depression, ` values from 2.33 to 6 for it to work properly though, not sure if 6 is actually the best value for that, but seems to allow it to fully look up/down without reversing everything at least.

Freecam feels a lot better to use now, there does still seem to be some jittering when looking up/down though, not sure if that's a problem with the game code, or maybe the way we update `fFreeCam_depression`...
(increasing the smoothing which earlier commit removed helps with the jittering, but would be sweet if we could fix it without needing the camera smoothing active, side-to-side camera movement doesn't seem to jitter at all..)

nipkownix:
Ah, pretty nice improvements!

> Sadly setting the smoothness too low makes camera movements pretty jerky, as explained at https://github.com/nipkownix/re4_tweaks/issues/436

I ended up hitting the same roadblock a few months ago. Was trying to make the mouse-look camera more responsive by removing this smoothing. Meh.

> Finally found a way to get rid of the camera orbiting, this took way too long to figure out, was about to try decompiling GC's debugCamera stuff but decided to try out zeroing the CAMERA_POINT stuff first, was surprised to see that did the trick 😸

Ohh! Nice work! I'm not at home for a couple of days, so I can't test this right now, but I'm looking forward to it.

emoose:
Figured a way to get rid of some of the sticking issues, took a while but think I got it calculating the m_target value mostly correct now (so now the g_transOfs stuff shouldn't be involved any more)

Still some jitteryness when looking up/down though, strange because left/right doesn't seem to have that issue, there is some code (@ 0x59E27A) that uses a `lr_rate.1171` variable which seems to be adding to a `target_aim_eff.1170` vec each frame, only seems to add the X axis though, for the others it's using the direction directly, maybe some kind of smoothing...
(if you disable that block by nopping 0x59E23F that does seem to make left/right start to jitter too, wonder if editing this stuff could help us get rid of the Y axis jitter in-game when smoothing is disabled...)

Also some weirdness when trying to look directly up/down, starts slowing down the rotation for some reason, don't really get why.. maybe there is some issue with whatever is using the m_target value, but not really sure.
For now I've just left it clamp to `-6 - 6`, should stop it just short of when it starts slowing the movement, means we can't look directly up/down anymore though...

emoose:
Rebased it onto latest master, there is one weird thing I noticed with this though, if you hold RMB while in freecam then moving the mouse will shift the camera position - that could actually be useful for lining up position instead of trying to tap WASD keys, but seems there's some problem with it which makes it move camera in a weird pattern.

I'm not sure how that's actually being done, doesn't seem we're checking for RMB / moving camera based on move movement anywhere...

nipkownix:
> E: think it's game doing it, key_btn_on gets KEY_FORWARD etc set when moving mouse forward with RMB pressed even without freecam enabled

Ah yeah, IIRC that has been a thing since forever. You can reproduce the same issue if you hold RMB and move the mouse while not having a weapon equipped. Probably yet another weird quirk in ``PadRead``.

nipkownix:
Just tested this - very nice!
Did notice some jerky movement if I looked down all the way, though. Maybe increasing the camera smoothness would help here?

emoose:
Do you mean it's jerky as you move it down, or is it when you pull it straight down and it seems to shift left/right a little at the end? I couldn't really find a good way to fix that, tried loading up an FPS with noclip mode and tried the same there and that also seemed to have the exact same issue, so I figured that's not a huge deal (then again I only checked with HL1 which is pretty old, maybe newer games might have fixed that, hm..)

Smoothing might be able to help with those, you can try changing the smoothing setting in Display while in freecam and see if it helps (thought about adding a seperate FreecamSmoothing option for that but idk, maybe just setting freecam to a low smoothing value would be good enough)

It seemed like the camera code does its own smoothing as well with the lr_rate stuff I mentioned, wonder if patching that out while keeping the CamSmooth stuff active might be any better.

nipkownix:
> or is it when you pull it straight down and it seems to shift left/right a little at the end?

This. Not a big deal though. Does HL1 really have the same issue? Never noticed it before :p

Another thing before we merge this: It was mentioned in #436 that the camera goes to Leon's feet when free cam is enabled.  Maybe there's a way to position it at Leon's eye level instead or something like that?

emoose:
Ah yeah was meaning to add something for that, guess we can just add some hardcoded Y/Z offset to the character position for now.

Not sure if we can keep the current camera rotation at all, will try playing around with it soon.

nipkownix:
Are we good to merge this as is? I don't think the camera rotation is that big of a deal, tbh. It already feels pretty nice with the Y change.

Issue #434, opened 2023-01-08
[feature request] change the handgun+silencer text


juuuulian-s:
### Describe the feature you'd like to see added

so in the base game, the examine text for the base handgun with the silencer attached describes it as being fully upgraded and having extra critical chance.
it's pretty clear that this was done to communicate to players that wesker's pistol was fully upgraded, the silencer being rather inconsequential in mercenaries.
now with the hd project adding a secret silencer obtainable in the main game, I think it's best if this examine text gets changed to something else so as not to mislead new players and such.

Mister-Curious:
Edit the MDT file for this weapon and you are good to go. 

https://residentevilmodding.boards.net/thread/9792/re4uhd-mdt-tool-tutorial

pas-de-2:
Good call, this description is confusing and makes the silencer seem very overpowered.

juuuulian-s:
apparently they can check for that and keep the original text for the fully upgraded pistol, leaving wesker's handgun examine text untouched. so, best of both worlds :)

albertre4HD:
If that's the case, that would be great! But while I'm writing these lines, something comes to my mind:... We'd need the translation in French, German, Italian, Spanish, Japanese, Chinese simplified and Chinese traditional, oops!!

PR #438, opened 2023-01-11
Misc: fix some item descriptions in the merchant and examine subscreens

pas-de-2:
https://github.com/nipkownix/re4_tweaks/issues/434

Workaround fix for [the silencer wrongly suggesting it gives the handgun a 5x crit chance](https://user-images.githubusercontent.com/116680301/211691071-2256ad09-a251-4f76-bc44-200203837815.jpg) when the handgun has less than max firepower:

While we're here, remove the infinite bullets description from the unupgraded Chicago Typewriter:

### DIFF
--- a/dllmain/Game.cpp
+++ b/dllmain/Game.cpp
@@ -1070,7 +1070,7 @@ bool re4t::init::Game()
 	pattern = hook::pattern("E8 ? ? ? ? B9 ? ? ? ? 83 C4 ? 66 3B ? 0F 84");
 	ReadCall(injector::GetBranchDestination(pattern.count(1).get(0).get<uint32_t>(0)).as_int(), WeaponId2ChargeNum);
 
-	// cItem__specialTuned funcptr
+	// cItem::specialTuned funcptr
 	pattern = hook::pattern("8B CE E8 ? ? ? ? 84 C0 74 ? C6");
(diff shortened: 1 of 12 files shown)


lextra2:
Nitpick, but why in the second image, is half the text box empty and it looks like it cuts to a new line right in the middle?

pas-de-2:
That's just how it's formatted in the MDT file.
```
{0x0000}  {0x1100}{0x5300}{0x0300}This machinegun is outfitted with a{0x0300}powerful .45 caliber magazine. {0x0800}{0x0100}
```
{0x0300} is a newline character.

It's mostly an artifact of the fact that the font was larger on the lower res GameCube, such that the sentence wouldn't fit on one line:

RB4E08_2023-01-10_23-55-10
We don't have the ability to edit the MDT data in memory yet. This fix is only redirecting what item it thinks is being looked at. In the future we could improve this to eg. restore the Handgun w/ Silencer text.

juuuulian-s:
does this mean some of the unlockables are getting a rebalance/proper upgrade progression?
it always bothered me how quickly you can make the mathilda and the handcannon busted op since every upgrade is available from the get go.

nipkownix:
Very clever workaround! Can't read the code right now, but I imagine you're not spoofing the IDs if the weapons are fully upgraded, right?

pas-de-2:
Yes, it checks for whether the weapon has its exclusive upgrade first, so eg. Wesker's pistol is unaffected.

nipkownix:
Many thanks!

PR #439, opened 2023-01-11
Added config option for RifleScreenShake recoil strength

pas-de-2:
I was playing with this more and felt the effect was a little too muted, so I added in a slider to configure the recoil strength.

Edit: seems the reason I felt it was muted is because `QuakeExec` is weaker at 60fps, as mentioned here https://github.com/nipkownix/re4_tweaks/issues/333#issuecomment-1334413970 I was originally testing it at 30fps. Going to try and see if I can't figure out a 60fps fix for that first.


pas-de-2:
I will close this as unnecessary given the 60fps fix. If someone requests it, can always add this later.

PR #440, opened 2023-01-12
FrameRateFixes: Add FixCameraShakeEffects

pas-de-2:
As mentioned here https://github.com/nipkownix/re4_tweaks/pull/439, I thought the rifle screenshake option seemed too muted until I realized from reading this https://github.com/nipkownix/re4_tweaks/issues/333#issuecomment-1334413970 comment that it was because playing at 60fps mutes the strength and duration of camera shake effects.

I tested this fix in 120fps at the r31c door and it looked good!

Edit: Hmm, after testing it more this doesn't seem like a proper fix. The camera is shakier at 60fps than it is at 30fps now. Too bad. I'll try to come up with a correct fix.


nipkownix:
Ah, just noticed this. Also tried doing something similar back when I posted that comment, but I never got it to look quite right either.

PR #446, opened 2023-01-21
HUD tweaks: added SmallerLifeMeter and SmallerActionPrompts

pas-de-2:
Added an option to shrink the life meter HUD that can be toggled ingame:

My goal was to approximate the scale of the GameCube HUD, before the widescreen ports made the HUD seem so big.
dolphin-s

I also added an option to shrink the action button prompts slightly to match:

Credit to albert's smaller HUD mod for helping me track down all the important textures.

I put these in a HUDTweaks.cpp, and came up with a slightly different fix for `bSideAlignHUD` that doesn't overwrite the life meter's x coordinate every frame. Can still be toggled ingame.


nipkownix:
That is incredible! Only nitpick I had was that the smaller HUD was positioned further away from the edges of the screen, making it look a bit offset. Pushed a commit to fix that.

Noticed an issue with the PRL though:
image

It fixes itself if I toggle ``SmallLifeMeter``, though. Not sure why.

pas-de-2:
Oh, that's because I moved the hook from `Cockpit::roomInit` to `Lifemeter::roomInit` at some point and forgot to re-test the PRL. Moved it to the end of `BulletInfo::roomInit` and that fixes it.

I prefer the lifemeter slightly offset at this resolution since it's so far away from the center of the screen. By shrinking it, you actually move it over to the right. I'll add a config option so the player can easily reposition it how they like.

nipkownix:
> Moved it to the end of BulletInfo::roomInit and that fixes it.

Ah, nice.

> I prefer the life meter slightly offset at this resolution since it's so far away from the center of the screen. I'll add a config option so the player can easily reposition it how they like.

Oh, didn't realize it was intentional. I'd appreciate the config option ^^

pas-de-2:
Should we set `re4t::cfg->HasUnsavedChanges` to false after writing to the ini?

nipkownix:
Huh, I thought we were already doing that? Might have missed it in the .ini reader PR though...

emoose:
`re4t_cfg::WriteSettings` func does have a `re4t::cfg->HasUnsavedChanges = false;` line at the end of it, but seems we return just before it gets ran:
https://github.com/nipkownix/re4_tweaks/blob/2ff10dd01a5ee460d979826f0b8cb6f68db078a6/dllmain/Settings.cpp#L916-L922

The block that holds that `return` will always be ran, but think that was meant to only be when the `trainerOnly` param was true, so saving trainer settings only would early return and not change the `re4t::cfg->HasUnsavedChanges` flag.
Easiest fix is probably just moving that `re4t::cfg->HasUnsavedChanges = false;` line into the `!trainerOnly` section.

Actually noticed that a few days ago trying out a static analyser against it, found a few other minor mistakes with it too, will probably make a PR for them soon.

PR #451, opened 2023-01-24
Misc minor fixes from static analyser

emoose:
Fixes some minor issues flagged by static analyser, hopefully all the fixes here should be good, but let me know if anything seems wrong.

There were two other things it flagged that I'm not sure about atm, again just minor things though:

Two Em names are defined for 0x13 in Game.cpp:
https://github.com/nipkownix/re4_tweaks/blob/2ff10dd01a5ee460d979826f0b8cb6f68db078a6/dllmain/Game.cpp#L179-L183
There's also no name for Em10 defined there, maybe 0x11 was meant to be 0x10, 0x12 was meant to be 0x11, and first 0x13 was meant to be 0x12? I don't really know what the names of each enemy is though so no idea if that's the case 😅 

and there's a weird check in QTEFixes.cpp of `!var || var`, so it passes every time, is something meant to be checked there?
https://github.com/nipkownix/re4_tweaks/blob/2ff10dd01a5ee460d979826f0b8cb6f68db078a6/dllmain/QTEFixes.cpp#L114

### DIFF
--- a/dllmain/AutoUpdater.cpp
+++ b/dllmain/AutoUpdater.cpp
@@ -264,7 +264,7 @@ void updateDownloadApply()
 			BOOL result_moveFile1 = MoveFileExW(new_module_path1.c_str(), new_module_path2.c_str(), MOVEFILE_REPLACE_EXISTING);
 			BOOL result_moveFile2 = MoveFileExW(new_ini_path1.c_str(), new_ini_path2.c_str(), MOVEFILE_REPLACE_EXISTING);
 			
-			if (!result_moveFile1 || !result_moveFile1)
+			if (!result_moveFile1 || !result_moveFile2)
 			{
 				spd::log()->info("{} -> Failed to rename file. Not critical, continuing...", __FUNCTION__);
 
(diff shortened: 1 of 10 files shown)


nipkownix:
> Two Em names are defined for 0x13 in Game.cpp:

Ah, there's only supposed to be ``{0x13, "Ganado / Merchant"}``, I must have duplicated that by accident. Removing the first one should be fine.

``if (re4t::cfg->bDisableQTE && (!re4t::cfg->bAutomaticMashingQTE || re4t::cfg->bAutomaticMashingQTE))`` is most likely supposed to just be ``if (re4t::cfg->bDisableQTE)``. I probably messed that one up back when I was renaming the vars :p
I'll revisit this soon, we can probably improve that prompt hook nowadays.

I'm a bit tired tonight, so I'll go through the changes tomorrow, but everything seems good at a first glance.

Which static analyzer are you using, btw? I should probably use it too :p

Thanks!

laymonage:
Might be worth adding the static analyzer to the CI setup, too. If the tool is easy enough to use I might be able to help set that up.

nipkownix:
Just took a look and everything seems good. Thanks again!

Issue #145, opened 2022-03-03
[Feature request] Controller/Gamepad - Analog progressive turning speed


80T:
In the light of an earlier suggestion https://github.com/nipkownix/re4_tweaks/issues/118 (Restored Camera Movement for Xinput Devices) I came to realize that there are some features already implemented in this game that could potentially combine to achieve a much more pleasant/satisfying experience while moving around in the game and that being without breaking the game's pacing, style or difficulty, but it would be more of a.. convenience/polish addition on top of what was already there.

So:
- I know in RE4 Tweaks there is a new mouse turning feature (with animations as well) that will allow anyone with a KB&M setup to turn smoothly, at any turning speed which is great
- I also know that with this port, there is a hard link between aiming and moving while using a controller/gamepad and while adjusting for one, it becomes detrimental to the other (either aiming with a huge deadzone, or have movement with highly sensitive turning)
- Another thing I know is that character turning with a gamepad/controller is not analog/progressive, but digital (like the Analog Stick is emulating the D-Pad - this is not an issue with the port, but how the original game always was.. but this could hopefully be changed on the PC.. ?)
- As the mouse turning on the KB&M setup makes total sense, on a controller/gamepad I can see how that would conflict with using the camera on the Right Analog "C-Stick" (as I have already tried achieving this goal using Steam Input custom configurations)

My suggestion here implies that with the help of RE4 Tweaks Hook/Mod (?) maybe we could take advantage of the "analog" progressive turn speed that is achieved by enabling "Mouse Turning" in RE4 Tweaks and have the Left stick "X" axis (turn Right/Left) to act as a simulated mouse input, (being already an analog input itself maybe it can be straight away passed "as is" as analog input, to.. do the progressive turning.. ?) while only keeping/passing the "Y" axis (forward/back movement) unaltered to the game, this way having no more issues with movement being to cardinal or ordinal biased, or too sensitive, etc. Also, we could have an adjustable sensitivity for turning (at least by altering the mouse sens)

[...]

PR #477, opened 2023-03-05
ControllerTweaks: added SmoothAnalogTurning

pas-de-2:
https://github.com/nipkownix/re4_tweaks/issues/145

This adds smooth analog turning for gamepad controls. It should feel very similar to the type A/B controls in RE5 and the 'Classic' controls in RE: Rev.

[smoothturn.webm](https://user-images.githubusercontent.com/116680301/222960655-54a665e4-6e65-43c1-9b6e-7fcbb941e893.webm)

Another major benefit to this option is that it's now possible to play with the deadzone slider set to 0 without struggling to keep Leon moving in a straight line.

[dinput8.zip](https://github.com/nipkownix/re4_tweaks/files/10891880/dinput8.zip)

Side note: I had to migrate some of the `bFixTurningSpeed` fixes here, which just left behind fixes for `pl_R1_Crouch`. Is that function actually used in the game anywhere, or is it a holdover from 3.5?

### DIFF
--- a/dllmain/ControllerTweaks.cpp
+++ b/dllmain/ControllerTweaks.cpp
@@ -121,12 +121,109 @@ void re4t::init::ControllerTweaks()
 		spd::log()->info("{} -> XInput deadzone changes applied", __FUNCTION__);
 	}
 
-	// Fix type III controls melee range firing bug
+	// Smooth analog turning
+	// makes left stick movement similar to the 'Classic' control schemes found in later REs
+	{
+		// pl_R1_Walk
+		auto pattern = hook::pattern("8B C1 83 E0 04 33 D2 0B C2 74 ? 8B ? ? ? ? ? D9");
+		static float* cPlayer__SPEED_WALK_TURN = *pattern.count(1).get(0).get<float*>(28);
+		struct WalkTurnHook
+		{
+			void operator()(injector::reg_pack& regs)
+			{
+				const float fDeadZone = 0.30f;
+
+				if (isController() && re4t::cfg->bSmoothAnalogTurning)
+				{
+					if (abs(*fAnalogLX) > fDeadZone)
+					{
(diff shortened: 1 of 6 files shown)


nipkownix:
Looks pretty good!

> which just left behind fixes for pl_R1_Crouch in FrameRateFixes.cpp. Is that function actually used in the game anywhere, or is it a holdover from 3.5?

Haven't checked if it is this one, but I think it is used _at least_ here:
https://youtu.be/KF9vYZuT1z8?t=902

Might be used in the mine cart section too, but I'm not sure.

> Add an option for RE: Revelations-style controls where stationary left/right turning is fully analog as well. Would need to figure out how to slow down the turn animation speed or it'll look really odd. 

Something like the Type B MouseTurning would work fine, right? I'm pretty sure the hooks for that would work if you changed them to use the controller stick values.

pas-de-2:
> Haven't checked if it is this one, but I think it is used _at least_ here:
> https://youtu.be/KF9vYZuT1z8?t=902

Ah good call, yes it's used to control the camera here.

> Something like the Type B MouseTurning would work fine, right?

That should work to physically turn Leon, but I'd like to find some way to slow down the animation too, so his feet are actually shuffling slower at 0.3 analog tilt vs 1.0. Something like this:

[output.webm](https://user-images.githubusercontent.com/116680301/225832294-6fbb0cca-3b9a-451c-b3f1-9ae44e8ba5dc.webm)

It's not too big a deal, though, I prefer RE5's controls anyways.

nipkownix:
> That should work to physically turn Leon, but I'd like to find some way to slow down the animation too, so his feet are actually shuffling slower at 0.3 analog tilt vs 1.0. Something like this:

Ah, right. I wanted to do something like that too, but couldn't find a way.

I'll go ahead and merge it, then. Many thanks, once again!

Issue #469, opened 2023-02-21
[feature request] option to disable continuous fire when the attack button is held


juuuulian-s:
### Describe the feature you'd like to see added

for pistols in particular, you can hold the fire button down and the gun will continue to fire until empty. im wondering if this can be avoided using a similar method to the one implemented for the matilda burststop.

additionally, is it possible to optionally disable reloading with the fire button when empty? forcing manual reloads and avoiding accidental ones (particularly useful when going for capacity upgrade refills)

PR #487, opened 2023-03-13
Added DisableAutomaticReload option

pas-de-2:
Fulfills one of the requests in https://github.com/nipkownix/re4_tweaks/issues/469

Test build:
[dinput8.zip](https://github.com/nipkownix/re4_tweaks/files/10961759/dinput8.zip)


nipkownix:
Nice! Have you had the chance to test every weapon with this enabled, btw?

pas-de-2:
Yeah, should be good.

juuuulian-s:
this feature has been so fun to play around with. i keep getting hit cuz i forgot to reload, its great!

pas-de-2:
Yeah, I think there's something satisfying about hearing the empty mag click sound and then having to make a decision. The auto reload takes that away, and sometimes even gets you in trouble.

PR #492, opened 2023-03-18
FixRetryLoadMouseSelector: crash fix

pas-de-2:
This option kept crashing<sup>[[log](https://github.com/nipkownix/re4_tweaks/files/11007378/crash.log)]</sup> for me whenever I'd press escape to skip the Hunnigan cutscene at the start of R101. Added a null check and the crashes stopped.


nipkownix:
Ah, hmm, I wonder why I've never seen this crash before. Thanks for catching it!

pas-de-2:
I was wondering similar, but I can reproduce the crash with the 1.7.7.6 DLL, so it doesn't seem to be the result of a recent regression elsewhere.

PR #497, opened 2023-03-19
Tweaked SmoothAnalogTurning

pas-de-2:
Small update for SmoothAnalogTurning after playtesting back and forth with RE5 some more, realized I needed to up the max turn speed so there's a smoother gradient to the standing left/right turn speed. Feels a lot better now.

Also, minor thing, but I wanted to make the starting difficulty modifiers from NTSC mode independent of restart just in case someone doesn't realize the option requires that when they first configure re4_tweaks and start a new game.


nipkownix:
Awesome, thanks for playtesting a bit more. I don't have a controller at the moment, so I couldn't test these changes myself.

PR #529, opened 2023-03-30
SmoothAnalogTurning: fixed acceleration curve

pas-de-2:
Did some more research and realized I needed to square the post-deadzone analog throw value to get a proper response curve when doing player rotation on an analog stick. *Now* it should feel right.

Test build:
[dinput8.zip](https://github.com/nipkownix/re4_tweaks/files/11107195/dinput8.zip)


nipkownix:
Sorry for taking so long. Thanks for the patches!

PR #632, opened 2025-07-12
Update Trainer.cpp to allow the player to enter doors even if Ashley is trapped

HumanYeeeee:
When the "Allow entering doors without Ashley" option is enabled, it doesn't account for Ashley triggering a trap and getting stuck. In such cases, the player would be unable to enter any doors. By disabling that flag (that is checked in CheckAshleyActive), it overrides that behavior, then the player can enter doors regardless. 

Test cases:
I tested this in the game when it's raining after encountering Ashley and bringing her back, in the village where the church is located and many places around it, there are many traps (and doors)

### DIFF
--- a/dllmain/Trainer.cpp
+++ b/dllmain/Trainer.cpp
@@ -975,10 +975,10 @@ void Trainer_Init()
 				regs.eax = (uint32_t)AshleyPtr();
 
 				// Make game check for Leon's position against itself, instead of checking against Ashley's position.
-				// Disable the flag that is set when Ashley is trapped
+				// Disables flags_STATUS_2_5024 which is set when Ashley is trapped
 				if (re4t::cfg->bTrainerAllowEnterDoorsWithoutAsh)
 				{	
-					FlagSet(GlobalPtr()->flags_STATUS_2_5024, uint32_t(Flags_STATUS::STA_PL_CHECK2), false);
+					FlagSet(GlobalPtr()->flags_STATUS_0_501C, uint32_t(Flags_STATUS::STA_SUB_CATCHED), false);
 					regs.eax = (uint32_t)PlayerPtr();
 				}
 			}
--- a/dllmain/Trainer.cpp
+++ b/dllmain/Trainer.cpp
@@ -978,7 +978,7 @@ void Trainer_Init()
 				// Disable the flag that is set when Ashley is trapped
 				if (re4t::cfg->bTrainerAllowEnterDoorsWithoutAsh)
 				{	
-					GlobalPtr()->flags_STATUS_2_5024[0] &= ~0x20000000;
+					FlagSet(GlobalPtr()->flags_STATUS_2_5024, uint32_t(Flags_STATUS::STA_PL_CHECK2), false);
 					regs.eax = (uint32_t)PlayerPtr();
 				}
 			}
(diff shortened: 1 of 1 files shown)


dllmain/Trainer.cpp
```diff
@@ -975,8 +975,12 @@ void Trainer_Init()
 				regs.eax = (uint32_t)AshleyPtr();
 
 				// Make game check for Leon's position against itself, instead of checking against Ashley's position.
+				// Disable the flag that is set when Ashley is trapped
 				if (re4t::cfg->bTrainerAllowEnterDoorsWithoutAsh)
+				{	
+					GlobalPtr()->flags_STATUS_2_5024[0] &= ~0x20000000;
```
emoose: `FlagSet(GlobalPtr()->flags_STATUS_0_501C, uint32_t(Flags_STATUS::STA_SUB_CATCHED), false);` might unset the same flag, haven't tested that though

HumanYeeeee: Yea it does work. It's also equivalent to:
```cpp
FlagSet(GlobalPtr()->flags_STATUS_2_5024, uint32_t(Flags_STATUS::STA_PL_CHECK2), false);
```
To be explicit about which flag would change. I will update the branch

emoose: STA_SUB_CATCHED should be the right flag afaik:
```STA_SUB_CATCHED = 66, // 0x20000000 (@ 0x8)```
The @ 0x8 there says FlagSet will add 0x8 to the flag ptr `flags_STATUS_0_501C`, so it'll be written to flags_STATUS_2_5024 instead (those flags_STATUS_X fields are really just a single array of ints that FlagSet accesses by `flagIdx / 32`, not actually sure why we left them as separate fields in the struct though)

HumanYeeeee: I was aware of that and debugged FlagSet for both statements, but I wasn't aware of the fact that those flags are meant to be elements of a flag array, even if both statements are equivalent it also dawned on me the name STA_SUB_CATCHED makes more sense, I suppose your FlagSet call makes a lot more sense but I should just add a comment to clarify which flag is modified since it's implicit

HumanYeeeee:
Hello, just wanted to check in if there is anything potentially missing or worth looking into, not sure this PR is of interest currently

laymonage (approved):
I haven't looked into the flags, but it makes sense to me. I rebased this PR on top of #637 and it works great. Thanks!

(I'm not a maintainer though, so I can't merge your PR. Just thought I'd leave a comment to report my testing.)

emoose:
Hm was looking into this some more, wondering if disabling `STA_SUB_CATCHED` could have more effects than intended.

Tried checking game to see where flag is checked but so far only found where it's being set/unset, and a check inside `sceAtFunc_save` which I think prevents saving if ashley is caught (which I guess this patch would affect, and let you save while shes trapped).

Not sure where the door check for the flag is happening yet, but if we could find that we could probably hook that and skip changing flags for it.

E: looks like it might be in `CheckAshleyActive`, that checks `Flags_STATUS_2_94[0] & 0x20000000` which is the STA_SUB_CATCHED flag, bunch of other code also seems to check same flag too though, maybe can just insert a hook in this func.

nipkownix:
Sounds like the hook would be the best way to do this. Guess I'll take a look soon.

PR #637, opened 2025-10-12
Update ModUtils to latest and fix CI configuration

laymonage:
Tried building the project and encountered the same issue as https://github.com/CookiePLMonster/ModUtils/issues/13. It's already fixed in ModUtils so we just need to update the submodule.

Note that due to https://github.com/CookiePLMonster/ModUtils/commit/b851d8b6da0a8144db4c2bfb62fb1099ff0f80a4, we need to update `PATCH_CALL` and `PATCH_JUMP` to `HookType:Call` and `HookType:Jump`, respectively.

I also updated and simplified the GitHub Actions workflow so that it works again, and it can now be [triggered manually on the GitHub UI](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow).


laymonage:
Apparently our pattern scanning no longer works after https://github.com/CookiePLMonster/ModUtils/commit/67b09f88c920f725d086ff582d345a879003e96e. Not sure how to fix it yet, but reverting that commit makes it work again.

emoose:
Ah guess that would probably break whatever data patterns we use, might be a way around that by passing the base address + SizeOfImage over though, but would probably need to update every `hook::pattern()` call that's checking for data :/

E: or wonder if we could make a wrapper around hook::pattern that passes that over, and change every call over to it? Finding the data-specific ones would be nicer though.

laymonage:
Thanks @emoose!

I pushed a commit that adds a wrapper function that uses ModUtil's scanning logic by default, with fallback to scanning other sections if it cannot find the pattern in the text section (as well as logging the section where the pattern is). You can also pass it a specific section so that it scans that pattern directly.

I used AI to generate the `SectionedPatterns.h` file and only did a quick review with my limited knowledge, so I'm sure it can use some improvements.

I've found that these patterns are in `.data`:

- https://github.com/nipkownix/re4_tweaks/blob/7cfcf5c4272f7d68d177661ffe84e9d1e0d2f383/dllmain/Gameplay.cpp#L312-L314
- https://github.com/nipkownix/re4_tweaks/blob/7cfcf5c4272f7d68d177661ffe84e9d1e0d2f383/dllmain/Misc.cpp#L1219-L1222
- https://github.com/nipkownix/re4_tweaks/blob/7cfcf5c4272f7d68d177661ffe84e9d1e0d2f383/dllmain/Misc.cpp#L1240-L1242
- https://github.com/nipkownix/re4_tweaks/blob/7cfcf5c4272f7d68d177661ffe84e9d1e0d2f383/dllmain/ToolMenu.cpp#L378-L379

and this one is in `.rdata`:
- https://github.com/nipkownix/re4_tweaks/blob/7cfcf5c4272f7d68d177661ffe84e9d1e0d2f383/dllmain/Misc.cpp#L1372-L1378

I looked into the `hook::pattern` calls and tried enabling all the options in the config to surface those patterns, but there might be more that I missed, especially if the scanning only happens while the game is running and with specific settings. But the game seems to be running fine so far with these changes.

I'm still learning the very basics of reverse engineering. I'm hoping we can bring more updates to this mod, as I think there are still quite a lot of people interested in OG RE4. If you have any suggestions so we can move this forward, that'd be great!

Thanks again for all your hard work!

laymonage:
Hello again!

It looks like ModUtils have been updated since the last time I worked on this. In particular, https://github.com/CookiePLMonster/ModUtils/commit/2e77fa1cf702d126ede5aba89511cdf74055feb4 made it so that `hook::pattern` scans all readable sections (based on [`IMAGE_SCN_MEM_READ`](https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-image_section_header#IMAGE_SCN_MEM_READ) flag). This means it can now scan non-code sections again, just like before.

That commit also provides utilities like `get_section_by_name`. I've reverted my AI-assisted commit that added the wrapper functions as it's no longer necessary.

However, I figured we might still be interested in marking the patterns I previously found to be in the non-code sections, and scan those specific sections to gain a small speedup (not sure how much). So I added a new commit that adds `Sections.h`/`Sections.cpp` to get those sections and store them in `re4t::sections` for easy reference, and use them for the non-code patterns. I learned a bit more C++, so it's not AI-generated this time :D

I've also updated the config to build the project in VS2026 now that the public beta for GitHub Actions is available: https://github.com/actions/runner-images/issues/13638

I think I'm happy with the current build setup. If you or nipkownix could get this in, that would be awesome. For now, I've merged this to the main branch of [my fork](https://github.com/laymonage/re4_tweaks) if anyone's interested. I'll start looking into the existing PRs and issues and see if I can get somewhere :)

nipkownix:
This looks pretty good. Thanks @laymonage!
I haven't been able to find time to work with this game in a while now, but I'd like to come back at some point.
@emoose, do you have merge permissions here? I think I enabled that for you some time ago, but I'm not sure. Feel free to merge this or do whatever else you think that would be a good change or addition. I've noticed some other nice PRs, but I won't be able to actually test anything on my end for a while.

emoose:
Just gave the action release build from fix/update-modutils a go (https://github.com/laymonage/re4_tweaks/actions/runs/24262134941), but the startup time is pretty long compared to the 1.9.1 build for me, probably takes 15-20 seconds now while 1.9.1 took maybe 2 or 3 :/

Don't really see much in the changes that could cause that though, wonder if new modutils might now be continuing the scan across whole PE for each hook instead of stopping scan at the first result?

E: ah yeah might be a ModUtils bug, in another project I had to switch away from using .count() and use .size() instead to fix some minor delays, didn't try looking into why that was needed then though.

Might have found fix for it, made a PR here: https://github.com/CookiePLMonster/ModUtils/pull/19

E2: played for a bit and PR seems fine besides that, once that fix lands happy to merge 👍

laymonage:
Thank you! I did notice the slowness, but I chalked it up to running it with debugging. Forgot to check with the release dll!

Out of curiosity, I tried adding a wrapper function that scans only the `text` section by default, and without the ModUtils fix it seems to have comparable speeds. When I tested it with the ModUtils fix, the startup time seems about the same, so the perf gain doesn't really stack up.

I'm not sure if it's worth doing considering the amount of find and replace needed. Not to mention the old ModUtils behaviour scanned the whole image and the performance was fine. It's here if you want to check it out though: https://github.com/laymonage/re4_tweaks/commit/1318f2438403db5f602775464cee14e6b84dd634

Also, just noticed that some of the actions in the CI config have got updates since I started working on this PR, and the ones we have now raise warnings about Node.js 20 being deprecated: https://github.com/nipkownix/re4_tweaks/actions/runs/24610389499

Should be simple to fix, I'll open a separate for it. Thanks again!!

Issue #300, opened 2022-07-30
Restore Echo effects and enable true Stereo Audio (maybe restore Dolby Pro Logic II too)


hyperspeedgx:
### Describe the feature you'd like to see added

In the GC and Wii version the game has echo/reverb in some areas inside the Castle, peharps this can be restored?

Also, PC ver. is know for be a mess in audio, there are more than echo/reverb effects missing, game runs at Mono audio quality while BGM seems to be the only thing playing at Stereo, some areas when you are playing the GC or Wii ver. you can hear much more sounds from the enviroment, like water sounds etc, if you do a playthough of GC or Wii with headphones you will notice even more, that's why some people still play the game in the emulator, audio is just at another level, can't say anything more.

btw: Dunno if GC had this but Wii and even the PS2 ver. has support to Dolby Pro Logic II, maybe this can be restored too?

albertre4HD:
It's true, certain effects lack its positionality in this port, but most of them still sound more to the left or to the right according to the sound source location.
Nevertheless, there's something broken/missing about the sound, yes.
The reverb sound is something I added to my "wishlist" a long ago, so, we are aware of that.
I really hope something can be done with that eventually! :D

sologuitarplaying:
Reloading and gunfire sound stronger and richer on GC and Wii as well. Whatever happened to the sound effects over the years of ports downgraded every sound effect in the game.

albertre4HD:
But the sound samples are the same. What's different is how they are treated by the game.

We are lucky this PC port has a good low-frequency sounds resample method. You only need to hear the ganados' voices and you'll clearly notice they sound better than in ps4/Xone/switch later ports, and they sound almost as good as in the GC. (positional audio is another story, I'm just talking about the perceived audio frequency, how clear or muffled they sound)

But as I said, how the sounds are resampled is what makes the difference. They are in all ports (but PS2) the same sound samples, some 32kz, some 22KZ, 11KZ, 8KZ...

Of course, higher frequency sounds will sound good in all ports, the difference is perceived with the lower frequency sounds. A good place to compare it is when Leon reaches the center of the Village and he uses the binoculars (most ganado's voices and certain enemies' sounds like the hayfork)

If the reloading and gunfire sound different, then it's probably because the game lacks some sound processing step --or however it's called-- the Gamecube did.

Maybe it's related to the missing reverb effect and it's all the same problem. Reverb effects add certain richness to the sounds, after all.

[...]

PR #650, opened 2026-04-21
Fix high-FPS sound-doubling by adding 30FPS/33ms timeslicing to SndCall

emoose:
One of main issues when playing above 60FPS (and probably above 30FPS) are certain sound effects being doubled, causing them to either increase volume or just get doubled entirely.
Footstep sounds are most noticeable but also affects some sounds from NPCs (picking up hay etc)

This seems to help those, sounds mostly match across 30FPS/60FPS/120FPS for me now, and gets rid of the weird reverb effect they had.
(wondering if this might help some audio issues that were reported at vanilla 60FPS as well #300, but not totally sure)

For this we just keep a static array tracking each calls `blk`/`call_no`/`pMod ptr`/return-addr, and time it was last played, each call is checked if it's already been played in last 33ms, if so then call is skipped, and we just return the sound-handle ID from the previous call.

Makes 120FPS a lot more viable to play now that footsteps don't sound like ass :P
Probably still a lot of audio issues with it though, didn't notice any major problems playing through for an hour, but always a chance it might break something.

---

**Release build download:** https://github.com/nipkownix/re4_tweaks/actions/runs/24751303372/artifacts/6567262546
You can switch to 120FPS+ by setting `UseDynamicFrametime` and `DisableFramelimiting` to true in dinput8.ini (but this might also help 60FPS too)

Notes:
  
- If you get crash loading into map at high FPS try restarting game, might have a fix for this soon (seems to be caused by them using separate audio thread without proper thread-safety... might be able to move most of that into main thread though)

- Some sounds are still triggered more frequently at higher FPS, eg. the cow breathing sounds, and maybe the Ganados talking to each other
  Looks like cows use RNG to decide if sound should be played, and since that RNG-check code runs at faster framerate, more of those checks occur and pass in the same amount of time.
Probably not an easy way to fix this besides hooking those funcs and adding a similar 33ms check (I'd guess 60FPS also has this issue too but 120FPS makes it way more noticeable.)

- This returns handle to the earliest trigger of the sound in the timeslice.
Might be code that checks handle to see if sound has finished in order to trigger something else, since we return an earlier instance of the sound, could cause things to trigger slightly earlier than they should (1-4 frames early)
Not sure if game ever does try checking sound handles though, only saw a single check in some menu func so far.
Tried making this mute the dupe sounds instead of skipping them but no luck with that yet :(

### DIFF
--- a/dllmain/AudioTweaks.cpp
+++ b/dllmain/AudioTweaks.cpp
@@ -109,19 +109,29 @@ bool __cdecl SndStop_Hook(uint32_t id, int time)
 		match = SndDedup::buffers[SndDedup::prev()].find(id);
 	}
 
-	if (match)
+	if (match && !match->isStopped)
 	{
-		if ((now - match->tick) < SndDedup::SLICE_DURATION)
+		if ((now - match->tick) < SndSlice::SLICE_DURATION) // Sound was started less than 33ms ago?
 		{
-			// Skip SndStop calls for sounds started less than 33ms ago.
 			// Some funcs use a `if (id) { SndStop(id); } id = SndCall(x)` pattern to start playing new sounds 
 			// At 30fps this is likely intended to allow interrupting an existing sound effect.
-			// Higher framerates can sometimes run this with the same SE multiple times within 33ms
+			// but higher framerates can sometimes run this with the same SE multiple times within 33ms
 			// Stopping and restarting the same sound before it can play, causing pops/cracks in the audio.
-			// 
-			// TODO: Possible this may cause some legit SndStop calls to become skipped, could be worth gating this check behind optional setting.
+			//
+			// It's possible some legitimate non-dupe SndStop calls might still happen in the 33ms window though
+			// So instead of stopping the sound here we'll set an expiry timer to stop it in next 33ms
+			// De-duped sounds will clear the timer inside SndCall, allowing it to continue playback, while Framelimiter_Hook will handle stopping any expired sounds.
+			{
+				match->expiryTick = match->tick + SndSlice::SLICE_DURATION;
+				match->sndStopParam = time;
+			}
+
 			return false;
 		}
+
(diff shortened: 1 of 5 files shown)


emoose:
Hm found one issue with this, some of the enemy shouts when they spot you aren't playing now

Looks like enemies actually use `SndStop` on their existing sound handle before they shout, so with our hook passing an earlier handle over, we're passing one that game had just stopped...

If I patch out the SndStop then it works fine (and looks like it fixes the audio pop at the start that albert mentioned before too 🤔)
Not a proper fix though.. maybe need to add some kind of ref-count to these dupe sounds instead.

(E: eh ref-count doesn't work since it's stopping before we can actually increment it, adding a flag to check if sound had been stopped before we de-dupe does fix it, but won't fix the audio pop issue sadly.. guess that needs a separate fix)

Updated build at https://github.com/nipkownix/re4_tweaks/actions/runs/24739426573/artifacts/6562349926

EDIT: added a commit that might help solve the pops/cracks too, but I'm pretty sure it could cause some other audio issues later in the game, going to test with that some more (links above don't include this commit)

emoose:
https://github.com/nipkownix/re4_tweaks/actions/runs/24751303372/artifacts/6567262546 should hopefully solve the pops/cracks now (even at 60FPS), played first act a couple times and didn't see any audio issues at 120FPS too 😸

If anyone tries it and notices any sounds that fail to play properly please let me know!

emoose:
Added the audio fixes from here into a new 120FPS build: https://github.com/nipkownix/re4_tweaks/pull/50#issuecomment-4308691060

nipkownix:
Awesome job!
@emoose, can this PR be merged without merging #50 ?

emoose:
Yeah this works without #50, it should be able to fix up a lot of the 60fps audio issues, might still need some more testing in the later areas of game though.

PR #656, opened 2026-07-15
Add configurable inventory item rotation hotkey

bingj1e:
## Problem

Keyboard and mouse users can rotate an inventory item's footprint through the game's native controls, but re4_tweaks does not expose that action as a configurable single-key binding in the F1 menu or INI.

The existing `FlipItemUp`, `FlipItemDown`, `FlipItemLeft`, and `FlipItemRight` bindings solve a different problem: they inject paired input masks (`0x00300000` or `0x00C00000`) that flip the 3D item model without changing its inventory footprint. Reusing a single shoulder-button bit in that hook still enters the model-flip path, so a 1x2 item remains 1x2.

## Change

- Add `HOTKEYS/RotateItem`, defaulting to `C`.
- Expose the binding in the F1 Hotkeys menu and make the section discoverable by searching for `rotate`.
- Resolve and call the game's existing inventory rotation routine when the key is pressed, so the footprint changes (for example, 1x2 to 2x1).
- Preserve the existing four model-flip bindings and behavior.
- Clearly distinguish footprint rotation from model-only flipping in the F1 menu, INI comments, and README.
- Use the existing `pzlPiece` SDK type for the native rotation call.

The binding rotates clockwise intentionally: repeated presses can reach every footprint orientation while keeping the change to one new default key.

## Impact

Keyboard and mouse players get one configurable key for clockwise footprint rotation, while the existing model-flip controls remain available and are now documented separately.

## Validation

- Built the current `master` diff as Release/Win32 with Visual Studio 2022, overriding the requested toolset to the locally installed `v143`; the full solution built successfully.
- Tested in game: pressing `C` while moving a 1x2 item changes it to 2x1.
- Verified the rotation signature has exactly one match in the Steam `bio4.exe` and resolves the native routine at `0x0040992B`.
- Confirmed the branch is based on current `master` (`1dd35a4`) and `git diff --check` passes.

The repository currently reports no automated checks for this branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,8 +63,9 @@ A project designed to fix and tweak the "UHD" port of Resident Evil 4 on [Steam]
  
  * **Allow reload without aiming** - Removes the need to be aiming the weapon before you can reload it. KB/M and Controller supported. (DInput untested). Also provides an extra option to disable the camera zoom when you reload weapons.
 
- * **Flip items in your inventory using the keyboard** - Creates key bindings to flip (not just rotate) items when playing with a keyboard and mouse, just like you can do if you play with a controller.
+ * **Rotate and flip items in your inventory using the keyboard** - Creates configurable keyboard and mouse bindings for two distinct actions: rotation changes an item's inventory footprint, while flipping changes only its 3D model orientation.
     * Default bindings: 
+		 * Rotate clockwise: C
 		 * Flip up: Home
 		 * Flip down: End
 		 * Flip left: Insert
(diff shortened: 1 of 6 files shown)


nipkownix:
So the idea here would basically be adding a way to re-bind the Delete and PageDown keys, from what I can see? I guess I'm not opposed to it, although I know you can already do this in the vanilla game by messing around with the input .ini file.

nipkownix:
Added pzlPiece_rotate() to Game.c/Game.h to follow what we already to with other game funcs.

Thank you!

Issue #645, opened 2026-04-01
BAD FIX MouseTurning!


eoegamer:
### Validation

- [x] Game has been updated to the latest RE4HD release (v1.1: https://www.re4hd.com/?p=9552)
- [x] Data has been [validated with SFV](https://www.re4hd.com/?p=9454) (a guide to using QuickSFV is available at that link, use that with this 1.1 SFV: [BIO4-HDProject1.1-SFV.zip](https://github.com/emoose/re4-research/files/9207370/BIO4-HDProject1.1-SFV.zip))

### Describe your issue here (drag+drop ZIP to attach it)

> `pInput->raw_mouse_delta`  works bad while enemy count > 100
> So here is the solution
```

	// X mouse delta adjustments
	{
		auto pattern = hook::pattern("A3 ? ? ? ? E8 ? ? ? ? A3 ? ? ? ? 8B 85 ? ? ? ? EB ? D9 C9");
		ptrMouseDeltaX = *pattern.count(1).get(0).get<uint32_t*>(1);
		struct MouseDeltaX
		{
			void operator()(injector::reg_pack& regs)
			{
				float deltaX = 0;
				//if (re4t::cfg->bUseRawMouseInput)
				//	deltaX = (pInput->raw_mouse_delta_x() / 10.0f) * g_MOUSE_SENS();
				//else
					deltaX = float(int(regs.eax));

				deltaX = deltaX * 0.5; // halve delta value to make it match the slower 60FPS speed that most people are used to

				*ptrMouseDeltaX = int32_t(deltaX / GlobalPtr()->deltaTime_70);
			}
		}; injector::MakeInline<MouseDeltaX>(pattern.count(1).get(0).get<uint32_t>(0), pattern.count(1).get(0).get<uint32_t>(5));
	}

	// Y mouse delta adjustments
	{
		auto pattern = hook::pattern("A3 ? ? ? ? E8 ? ? ? ? A3 ? ? ? ? 8B 85 ? ? ? ? EB ? E8");
		ptrMouseDeltaY = *pattern.count(1).get(0).get<uint32_t*>(1);
		struct MouseDeltaY
		{
			void operator()(injector::reg_pack& regs)
			{
				float deltaY = 0;
				//if (re4t::cfg->bUseRawMouseInput)
				//	deltaY = -((pInput->raw_mouse_delta_y() / 7.0f) * g_MOUSE_SENS());
				//else
					deltaY = float(int(regs.eax));

[...]

PR #659, opened 2026-08-16
Fix raw mouse delta being multiplied during slow frames

### LINKED ISSUE #460: Raw Mouse Input problem
### Validation

- [X] Game has been updated to the latest RE4HD release (v1.1: https://www.re4hd.com/?p=9552)
- [X] Data has been [validated with SFV](https://www.re4hd.com/?p=9454) (a guide to using QuickSFV is available at that link, use that with this 1.1 SFV: [BIO4-HDProject1.1-SFV.zip](https://github.com/emoose/re4-research/files/9207370/BIO4-HDProject1.1-SFV.zip))

### Describe your issue here (drag+drop ZIP to attach it)

I noticed if have many enemies on screen if have the RawMouseInput option enable, Leon aim don't work well. The sensibility get compleatly broken and rotate camera too.

mthsk:
Fixes #460
Supersedes the workaround proposed in #645

**Problem**
Raw mouse input accumulated from `WM_INPUT` was stored in a single buffer whose lifetime was tied to `next_frame()`. Gameplay hooks could read the same accumulated delta more than once, while input arriving after gameplay had already read the buffer could instead be cleared by `next_frame()` before the next gameplay update consumed it.

Under load, this made raw mouse aiming inconsistent and noticeably choppy even when the game remained close to its target frame rate.

**Fix**
* Maintain a separate `_raw_mouse_game_delta` buffer dedicated to gameplay.
* Provide `consume_raw_mouse_delta_x()` and `consume_raw_mouse_delta_y()`, which read the current value under lock and immediately zero it.
* Update the X and Y gameplay hooks in `KeyboardMouseTweaks.cpp` to call the consume variants.
* Leave the existing `raw_mouse_delta_x/y` buffer untouched so menus, cutscenes, and any other non-gameplay consumers continue to behave as before.
* Lock consumption against concurrent raw-input writes using the existing `_mutex`.
* **Do not** clear `_raw_mouse_game_delta` in `next_frame()`, only explicit consumption or `clear_raw_mouse_delta()` should zero it, preventing unconsumed gameplay delta from being discarded at the frame boundary.

**Before:** (Aiming is choppy and unresponsive, especially in room r208)

https://github.com/user-attachments/assets/2ebfc61f-f03a-417c-8767-516e896fc6b5

**After:** (Aiming is smooth and responsive)

https://github.com/user-attachments/assets/505372bf-9bcf-4f72-a370-5cbb9059868c

### DIFF
--- a/dllmain/KeyboardMouseTweaks.cpp
+++ b/dllmain/KeyboardMouseTweaks.cpp
@@ -186,7 +186,11 @@ void re4t::init::KeyboardMouseTweaks()
 			{
 				double deltaX = 0;
 				if (re4t::cfg->bUseRawMouseInput)
-					deltaX = (pInput->raw_mouse_delta_x() / 10.0f) * g_MOUSE_SENS();
+				{
+					// The game can run more than one update while recovering from a slow frame.
+					// Consume the movement so those updates cannot apply the same delta repeatedly.
+					deltaX = (pInput->consume_raw_mouse_delta_x() / 10.0f) * g_MOUSE_SENS();
+				}
 				else
 					deltaX = double(int(regs.eax));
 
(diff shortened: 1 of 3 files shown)


nipkownix:
Looks good. Thank you!

